PR #13, opened 2026-02-16
Load testing

agatx:
- Added conduit for WS load testing
- Added internal stats monitoring
- Tuned parameters for higher concurrency


agatx:
Addressed all review feedback in commit `3d05218`:

- Added `80:80` back to production nginx publish list in `docker-compose.prod.yml`.
- Hardened `/api/internal/stats`: when enabled, token is now required; missing token config returns `503`, missing/mismatch header returns `401` (plus new tests in `server/internal_stats_test.go`).
- Fixed load ramp cancellation control flow in `runner.go` (labeled break + explicit canceled-step return).
- Moved initial stats baseline fetch to after stabilization to avoid cross-step delta contamination.
- Fixed join cancellation path in `client.go` to mark expected close and close the socket on `ctx.Done()`.
- Reworked expected-close tracking to be connection-generation-aware, avoiding false unexpected-disconnect counts during reconnect storms.
- Corrected TX metrics accounting to increment only on successful enqueue in `signaling.go`.
- Updated loadtest runner/docs for mandatory internal stats token when stats are enabled.

Validation:
- `go test ./...` (server module) passes.
- shell syntax checks pass for `deploy.sh`, `client/public/tools/setup.sh`, and `server/loadtest/run-local.sh`.

PR #75, opened 2026-04-27
Phase 2 partial: foreground / Doze force-ping (#8)

agatx:
When iOS suspends a backgrounded process or Android puts it in Doze, the
WebSocket can be silently killed without the SDK noticing. Today the only
detection is the regular ping cycle, so the SDK reports `connected` for
up to `pingIntervalMs = 12 s` after the app comes back. This adds an
explicit foreground hook that issues a synthetic ping and arms a 2 s
deadline, force-closing the transport on miss so the existing reconnect
path runs immediately.

- Shared `foregroundForcePingTimeoutMs = 2_000` constant lands in all
  three SDKs and the parity check.
- iOS subscribes to `UIApplication.willEnterForeground` /
  `didEnterBackground`; Android tracks Activity start/stop counts via
  `Application.ActivityLifecycleCallbacks` (no new dep on
  `lifecycle-process`).
- `SignalingProvider.forceReconnectIfStale(timeoutMs:)` is the new
  public hook (default no-op); `SessionSignaling.forcePingWithDeadline`
  is the internal one.
- `SignalingClient` uses a monotonic `pongSeq` counter so the deadline
  arms close only when no pong arrives between ping and timeout —
  robust against same-ms test clocks.

Tests: 6 iOS + 6 Android cover ping emission, deadline-miss close,
pong-arrival cancel, no-op when disconnected, repeated calls cancel
prior deadlines, and post-close calls stay no-op.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/client-android/serenada-core/src/main/java/app/serenada/core/SerenadaServerProvider.kt
+++ b/client-android/serenada-core/src/main/java/app/serenada/core/SerenadaServerProvider.kt
@@ -119,6 +119,10 @@ internal class SerenadaServerProvider(
         sendRawMessage(type = type, payload = payload)
     }
 
+    override fun forceReconnectIfStale(timeoutMs: Long) {
+        signaling.forcePingWithDeadline(timeoutMs)
+    }
+
     override suspend fun getIceServers(): List<PeerConnection.IceServer> {
         val token = currentTurnToken?.takeIf { it.isNotBlank() } ?: return emptyList()
         return suspendCancellableCoroutine { continuation ->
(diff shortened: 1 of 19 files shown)


agatx:
@codex please review

PR #80, opened 2026-04-29
Resilience #1: dispatch + consume negotiation_dirty / relay_failed

agatx:
## Summary

Activates the Phase 1 server-side dirty-pair protocol on all three SDKs. Closes failure-mode #1 in [docs/resilience-failure-modes.md](docs/resilience-failure-modes.md). Builds on the #4 snapshot gate that landed in [#79](https://github.com/agatx/serenada/pull/79) — the server emits \`negotiation_dirty{with: cid}\` to active peers right after the post-reconnect snapshot, telling them which specific peer needs fresh renegotiation.

- **Behavior change**: when a previously-suspended peer reattaches and the SDK had pending negotiation traffic to it during the suspension, the SDK now schedules a glare-safe ICE restart for *just that peer* (rather than waiting forever for an answer that will never come).
- **\`relay_failed\` is informational only**: surfaced as a logged event but not directly acted on. The same dirty-pair condition will surface as \`negotiation_dirty\` once the suspended target reattaches; acting on both would double-trigger ICE restart. A future slice can optimize the in-flight offer-timeout for suspended targets.

## Per-platform shape

| Platform | Provider event | Consumer | Test file |
|---|---|---|---|
| Web | \`negotiationDirty\` / \`relayFailed\` events on \`SignalingProvider\` | \`MediaEngine.scheduleDirtyPairRestart(cid)\` | [SerenadaSession.test.ts](client/packages/core/test/SerenadaSession.test.ts) |
| Android | \`Listener.onNegotiationDirty\` / \`onRelayFailed\` | \`peerNegotiationEngine.scheduleIceRestart(cid, ...)\` | [SessionDirtyPairTest.kt](client-android/serenada-core/src/test/java/app/serenada/core/SessionDirtyPairTest.kt) |
| iOS | \`signalingProviderDidReceiveNegotiationDirty\` / \`...RelayFailed\` | \`peerNegotiationEngine.scheduleIceRestart(remoteCid:...)\` | [SessionDirtyPairTests.swift](client-ios/SerenadaCore/Tests/SerenadaCoreTests/SessionDirtyPairTests.swift) |

Each platform covers three cases: \`negotiation_dirty\` schedules ICE restart for the named peer; unknown CID is a no-op; \`relay_failed\` is logged but does not trigger ICE restart.

## Why route through existing per-CID \`scheduleIceRestart\`

All three platforms already have a per-peer ICE-restart entry that goes through the canonical glare/cooldown guards:
- \`isMakingOffer\` glare protection (perfect-negotiation pattern)
- \`ICE_RESTART_COOLDOWN_MS = 10_000\` cooldown
- \`canOffer\` checks (signaling connected, slot ready, host arbitration)

Calling that path means \`negotiation_dirty\` gets the same robustness as every other ICE restart trigger today.

## Test plan

- [x] \`scripts/check-resilience-constants.mjs\` (20 constants ✓)
- [x] \`scripts/check-version-parity.mjs\` (8 sources @ 0.6.2 ✓)
- [x] \`server/go test ./...\` ✓
- [x] Web \`vitest run\` (312 tests, was 310) ✓
- [x] Android \`:serenada-core:testDebugUnitTest\` ✓
- [x] iOS \`xcodebuild test\` ✓
- [ ] Manual: A drops mid-call, B sends offer to A while A is suspended → server replies \`relay_failed\`; A reattaches → server sends \`negotiation_dirty\` to B → B's ICE restart converges in one round

### DIFF
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,0 +1,1 @@
+{"sessionId":"2a1e8903-08ff-4306-af3d-207716cb0439","pid":12128,"procStart":"Wed Apr 29 03:19:58 2026","acquiredAt":1777441942904}
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,1 +1,0 @@
-{"sessionId":"2a1e8903-08ff-4306-af3d-207716cb0439","pid":12128,"procStart":"Wed Apr 29 03:19:58 2026","acquiredAt":1777441942904}
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ secrets/
 /.webrtc-ios-build
 client-ios/Resources/GoogleService-Info.plist
 /.claude/worktrees
+/.claude/scheduled_tasks.lock
 /tools/smoke-test/artifacts
 
 # ralphex progress logs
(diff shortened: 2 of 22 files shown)


agatx:
@codex please review

PR #105, opened 2026-05-21
Android: opt-in Frontline call UI variant

agatx:
## Summary
- New `SerenadaCallUiVariant.Frontline` opts the prebuilt Android call UI into an audio-first layout tuned for frontline workflows (large controls, world-camera-first, More sheet, multi-party tiles). Standard variant unchanged.
- URL-first Frontline sessions are configured `defaultVideoEnabled = false` with camera order `world → selfie → composite`; session-first hosts configure the same on `SerenadaConfig`.
- Host app exposes a Settings toggle (Standard / Frontline) wired through `SettingsStore.callUiVariant` and `CallManager.updateCallUiVariant`.
- Adds `CallState.callStartedAtMs` (seeded at join, refined from server-reported `participant.joinedAt` when plausible) for the Frontline elapsed-time chip.
- Shared helpers consolidated: `initialsFor` reused from `RemoteAvatar`, `shareLink` promoted to `internal`, no parallel utilities.

## Codex review fixes
- Render `layout.localPip` in multi-party Frontline grid mode; previously the GRID renderer iterated only `layout.tiles` and the local self-view disappeared in 3+-participant calls.
- Drop `config.videoEnabled` / `config.uiVariant` from `SerenadaCallFlow`'s session `remember` key. Re-keying spawned a second owned session without disposing the first when these toggled mid-call.

## Test plan
- [x] `./gradlew :serenada-core:testDebugUnitTest :app:testDebugUnitTest` — passes (new `callStartedAtMs` + `SettingsStore.normalizeCallUiVariant` coverage)
- [x] Standard variant: 1:1 join, mute, video toggle, flip camera, screen share, end call (no regression)
- [x] Frontline 1:1: audio-first join, share-link from waiting state, world↔selfie↔composite cycle, end call
- [x] Frontline 3-party: confirm local self-view tile is rendered alongside remote grid tiles
- [x] Toggle Standard ↔ Frontline in Settings between calls; subsequent call uses the new variant; no orphan session log entries
- [x] Verify Frontline waiting screen does not show the standard QR code


agatx:
@codex please review

klavaotkrytova (changes requested):
Thanks for the fixes. I found one compile-blocking issue before I could get into runtime behavior:

- `client-android/serenada-call-ui/src/main/java/app/serenada/callui/FrontlineCallScreen.kt:259-260` passes `chooserTitle` twice to `shareLink(...)`. Kotlin rejects duplicate named arguments, so the Android build will fail until one of these lines is removed.

Small hygiene note: `git diff --check origin/main...HEAD` also reports a trailing blank line at EOF in `FrontlineCallScreen.kt`.

I could not run the Gradle tests locally because this checkout environment has no Java configured (`JAVA_HOME is not set and no java command could be found in your PATH`).

klavaotkrytova (approved):
Follow-up after fixing my local Android setup: I reran the requested test target and it passes locally.

`cd client-android && JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64 ./gradlew :serenada-core:testDebugUnitTest :app:testDebugUnitTest`

Result: BUILD SUCCESSFUL in 39s.

I also rechecked `FrontlineCallScreen.kt`; the earlier duplicate `chooserTitle` note was from my inspection output, not the actual file. Sorry for the false alarm. The only remaining nit I see is `git diff --check` reporting a trailing blank line at EOF in `FrontlineCallScreen.kt`, which is non-blocking.

PR #116, opened 2026-05-27
Fix video calls stuck in "Reconnecting" (renegotiation races)

agatx:
## Summary

Two test devices got stuck in **"Reconnecting"** on 1:1 video calls with Serenada 0.7.0. Two independent bugs in the new `PeerNegotiationEngine` flow combined so the call never connected. This fixes both, across Android, iOS, and web.

## Root cause

On call start the offerer sends its first offer, then a `pending-retry` ICE restart immediately fires a second offer. Two things go wrong:

1. **offerId clobber (commit 1).** The answer-success path cleared the pending local offerId unconditionally. The renegotiation offer reassigns that id before the first answer's `setRemoteDescription` callback runs, so the clear clobbers the *new* offer's id, and the new offer's answer is then dropped as a `stale offerId`. ICE tears down with no completed negotiation → permanent reconnect loop.
2. **spurious pending-retry (commit 2).** That second offer should never have fired. `registerNetworkCallback` (Android) / `NWPathMonitor.start()` (iOS) replay the current network as an initial `onAvailable` / `.satisfied` callback right after registration. The session treated that baseline as a handover and scheduled a `network-online` ICE restart mid-first-offer. Deferred while `HAVE_LOCAL_OFFER`, it replayed as `pending-retry` the instant the first answer landed.

Bug 1 made bug 2 fatal. With both fixed, the call connects with a single offer/answer.

## Fixes

- **Drop the offerId clobber** — only clear the pending entry when it still matches the offerId we completed. (`PeerNegotiationEngine.kt` / `.swift`, `MediaEngine.ts`)
- **Gate the `network-online` restart** behind a `hasEverConnectedPeer` latch (set on first `CONNECTED`, reset per call) so the registration-replay baseline can't trigger it. (`SerenadaSession.kt` / `.swift`)

Web needs only the first fix: `window 'online'` / `connection 'change'` fire on real transitions, not on registration, so it has no spurious startup restart.

## Verification

- **Android** — built locally, installed on two devices. Before: stuck Reconnecting. After fix 1 only: connects but with a wasteful extra renegotiation. After both: single offer/answer, both peers reach `CONNECTED`, **0 stale-offer drops, 0 ICE failures** across repeated calls.
- **Web** — `tsc --noEmit` clean; all **368 core tests pass** (incl. 23 `MediaEngine` tests).
- **iOS** — edits applied for parity but **not compiled or device-tested** in this environment. Please build + smoke a call before merge.

## Out of scope

Separate, pre-existing issue observed in the same logs: LAN-direct envelope sends fail and fall back to relay (adds latency, not a call blocker).

### DIFF
--- a/client-android/serenada-core/src/main/java/app/serenada/core/SerenadaSession.kt
+++ b/client-android/serenada-core/src/main/java/app/serenada/core/SerenadaSession.kt
@@ -142,12 +142,22 @@ class SerenadaSession internal constructor(
     /** Whether the microphone is muted specifically because external audio, such as push-to-talk, is active. */
     val isMicMutedByExternalAudio: StateFlow<Boolean> = _isMicMutedByExternalAudio.asStateFlow()
 
+    // Latched true once the call's media has connected at least once. A "network-online"
+    // ICE restart is a recovery mechanism for an established call; firing it during initial
+    // setup is never useful and is actively harmful: registerNetworkCallback replays
+    // onAvailable for every already-connected network right after registration, and that
+    // replay would otherwise schedule an ICE restart mid-first-offer, forcing a redundant
+    // renegotiation the moment the first answer lands (observed as a "pending-retry").
+    private var hasEverConnectedPeer = false
+
     private val networkCallback = object : ConnectivityManager.NetworkCallback() {
         override fun onAvailable(network: Network) {
             handler.post {
                 if (_state.value.phase == CallPhase.InCall) {
                     if (isConnectionDegraded()) markConnectionDegraded()
-                    peerNegotiationEngine.scheduleIceRestart("network-online", 0)
+                    if (hasEverConnectedPeer) {
+                        peerNegotiationEngine.scheduleIceRestart("network-online", 0)
+                    }
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


agatx:
Addressed the review feedback in 85c0060a, on all three platforms (Android, iOS, web):

**Legacy/no-offerId regression (Copilot, all platforms).** The completed offer id is now the pending id captured before `setRemoteDescription` (`pendingOfferId ?? offerId`), not the answer's `offerId`. In the `LEGACY_OFFER_ID` case `pendingLocalOfferId` holds a real generated id, so the previous `== offerId` guard never cleared it and `currentNegotiationId` was set to `__legacy__`, tagging later outgoing ICE with a stale id. The real pending id is now cleared and recorded.

**Per-peer vs per-offer state on the retry race (Codex, P2).** `clearOfferTimeout` / `clearPendingIceRestart` (and the id / `currentNegotiationId` updates) now run only while the slot's pending offer still equals the completed one. If a newer renegotiation offer replaced it during the async answer-apply, we no longer cancel that offer's timeout / pending-retry, so it can't get stranded in `have-local-offer` if its answer is lost.

Both reduce to one change: finalize negotiation state against the *captured, validated* offer, and only while it's still the pending one.

Verification: Android `serenada-core` compiles; web `tsc --noEmit` clean + 368 core tests pass (incl. `MediaEngine`); iOS mirrors the same logic (not built in this environment). The original Android on-device repro still connects with a single offer/answer.

PR #118, opened 2026-05-31
[codex] Add mobile system picture in picture

agatx:
## Summary
- Add mobile system Picture in Picture support for Android and iOS call UI integrations.
- Keep PiP content simple: selected video feed when available, otherwise a scaled avatar/initials placeholder.
- Add host-app hooks, background camera handling on iOS, Android PiP activity behavior, and SDK integration docs.

## Notes
- Android PiP intentionally omits call controls and exits PiP when the call ends.
- iOS uses the system video-call PiP path and maps restore animation geometry to the active large call surface or no-video avatar cluster.

## Validation
- `git diff --check`
- `xcodegen generate`
- `xcodebuild -project SerenadaiOS.xcodeproj -scheme SerenadaiOS -destination 'id=CA6FFD90-98C3-4580-BA33-11A912FC0BA9' build`
- `./scripts/deploy_to_device.sh --udid 9C3F87BA-CA63-58AC-8A17-076E7CFFDBFE`
- `./gradlew assembleDebug`

### DIFF
--- a/client-android/serenada-call-ui/src/main/java/app/serenada/callui/CallScreen.kt
+++ b/client-android/serenada-call-ui/src/main/java/app/serenada/callui/CallScreen.kt
@@ -260,13 +260,8 @@ internal fun CallScreen(
             )
         }
 
-    val isCallChromeVisible = !isSystemPictureInPicture
-    val areControlsActuallyVisible = areControlsVisible && isCallChromeVisible
-
     val toggleControlsVisibility: () -> Unit = {
-        if (isSystemPictureInPicture) {
-            Unit
-        } else if (areControlsVisible) {
+        if (areControlsVisible) {
             areControlsVisible = false
             wereControlsLastHiddenByAutoHide = false
         } else {
@@ -280,7 +275,7 @@ internal fun CallScreen(
 
     // Auto-hide controls
     LaunchedEffect(areControlsVisible, uiState.phase, isControlsAutoHideEnabled) {
-        if (areControlsActuallyVisible && uiState.phase == CallPhase.InCall && isControlsAutoHideEnabled) {
+        if (areControlsVisible && uiState.phase == CallPhase.InCall && isControlsAutoHideEnabled) {
             delay(8000)
             wereControlsLastHiddenByAutoHide = true
             areControlsVisible = false
@@ -308,11 +303,15 @@ internal fun CallScreen(
     SerenadaTheme(theme) {
       androidx.compose.runtime.CompositionLocalProvider(LocalAvatarCache provides avatarCache) {
         if (isSystemPictureInPicture) {
+            val systemPipRemoteCid =
+                pinnedParticipantId?.takeIf { pinned ->
+                    uiState.remoteParticipants.any { participant -> participant.cid == pinned }
+                } ?: uiState.remoteParticipants.firstOrNull()?.cid
             val feed =
                 selectSystemPictureInPictureFeed(
                     localIsLarge = effectiveLocalLarge,
                     localVideoEnabled = uiState.localVideoEnabled,
-                    hasRemote = uiState.remoteParticipants.isNotEmpty(),
+                    remoteCid = systemPipRemoteCid,
                 )
             SystemPictureInPictureContent(
                 uiState = uiState,
@@ -322,8 +321,8 @@ internal fun CallScreen(
                 remoteContentScale = if (remoteVideoFitCover) ContentScale.Crop else ContentScale.Fit,
                 attachLocalSink = attachLocalSink,
                 detachLocalSink = detachLocalSink,
-                attachRemoteSink = attachRemoteSink,
-                detachRemoteSink = detachRemoteSink,
+                attachRemoteSinkForCid = attachRemoteSinkForCid,
+                detachRemoteSinkForCid = detachRemoteSinkForCid,
             )
         } else {
         BoxWithConstraints(
@@ -353,7 +352,7 @@ internal fun CallScreen(
                         uiState.connectionState == "CONNECTED")
         val animatedPipBottomPadding by
         animateDpAsState(
-            targetValue = if (areControlsActuallyVisible) 130.dp else 48.dp,
+            targetValue = if (areControlsVisible) 130.dp else 48.dp,
             animationSpec = tween(durationMillis = controlsAnimationDuration),
             label = "pip_bottom_padding"
         )
@@ -366,7 +365,7 @@ internal fun CallScreen(
         val pipInnerCornerRadius =
             if (pipCornerRadius > pipContentPadding) pipCornerRadius - pipContentPadding else 0.dp
         val mainVideoBottomPadding by animateDpAsState(
-            targetValue = if (areControlsActuallyVisible) 134.dp else 0.dp,
+            targetValue = if (areControlsVisible) 134.dp else 0.dp,
             animationSpec = tween(durationMillis = controlsAnimationDuration),
             label = "main_video_bottom_padding"
         )
@@ -637,7 +636,7 @@ internal fun CallScreen(
         }
 
         // Debug tap target — only when debug overlay is enabled via config
-        if (config.debugOverlayEnabled && isCallChromeVisible) {
+        if (config.debugOverlayEnabled) {
             Box(
                 modifier =
                     Modifier.align(Alignment.TopStart)
@@ -674,7 +673,7 @@ internal fun CallScreen(
         }
 
         // Waiting State Overlay
-        if (uiState.phase == CallPhase.Waiting && !effectiveLocalLarge && isCallChromeVisible) {
+        if (uiState.phase == CallPhase.Waiting && !effectiveLocalLarge) {
             WaitingOverlay(
                 roomShareUrl = roomShareUrl,
                 onInviteToRoom = onInviteToRoom,
@@ -686,7 +685,7 @@ internal fun CallScreen(
 
         // Reconnecting Indicator
         AnimatedVisibility(
-            visible = showReconnectingBadge && isCallChromeVisible,
+            visible = showReconnectingBadge,
             enter = fadeIn(),
             exit = fadeOut(),
             modifier = Modifier.align(Alignment.TopCenter).padding(top = 64.dp)
@@ -745,9 +744,9 @@ internal fun CallScreen(
         }
         val snapshotHandler = onSnapshotRequested
         val showSnapshotButton =
-            snapshotSource != null && snapshotHandler != null && areControlsActuallyVisible
+            snapshotSource != null && snapshotHandler != null && areControlsVisible
 
-        if (isCallChromeVisible && (showFlashButton || showRemoteFitButton || showSnapshotButton)) {
+        if (showFlashButton || showRemoteFitButton || showSnapshotButton) {
             val configuration = androidx.compose.ui.platform.LocalConfiguration.current
             val isLandscape = configuration.orientation ==
                 android.content.res.Configuration.ORIENTATION_LANDSCAPE
@@ -797,18 +796,22 @@ internal fun CallScreen(
             }
 
             val snapshotIcon: @Composable () -> Unit = {
-                if (showSnapshotButton && snapshotHandler != null && snapshotSource != null) {
-                    IconButton(
-                        onClick = { snapshotHandler(snapshotSource) },
-                        modifier = Modifier.size(44.dp)
-                            .background(Color.Black.copy(alpha = 0.4f), CircleShape)
-                            .testTag("call.takeSnapshot")
-                    ) {
-                        Icon(
-                            imageVector = Icons.Default.PhotoCamera,
-                            contentDescription = resolveString(SerenadaString.CallTakeSnapshot, strings),
-                            tint = Color.White
-                        )
+                if (areControlsVisible) {
+                    snapshotHandler?.let { handler ->
+                        snapshotSource?.let { source ->
+                            IconButton(
+                                onClick = { handler(source) },
+                                modifier = Modifier.size(44.dp)
+                                    .background(Color.Black.copy(alpha = 0.4f), CircleShape)
+                                    .testTag("call.takeSnapshot")
+                            ) {
+                                Icon(
+                                    imageVector = Icons.Default.PhotoCamera,
+                                    contentDescription = resolveString(SerenadaString.CallTakeSnapshot, strings),
+                                    tint = Color.White
+                                )
+                            }
+                        }
                     }
                 }
             }
@@ -838,7 +841,7 @@ internal fun CallScreen(
 
         // Controls Bar
         AnimatedVisibility(
-            visible = areControlsActuallyVisible,
+            visible = areControlsVisible,
             enter =
                 fadeIn(animationSpec = tween(durationMillis = controlsAnimationDuration)) +
                         slideInVertically(
@@ -1540,7 +1543,7 @@ private fun MultiPartyStage(
                 val topChromePx = with(density) { 20.dp.toPx() }
                 val bottomChromePx = with(density) { (bottomPadding + 4.dp).toPx() }
 
-                if (useComputedLayout && localCid != null) {
+                if (useComputedLayout) {
                 // Focus/content mode: use computeLayout for primary + filmstrip rendering
                 val contentSource = if (hasLocalContent) {
                     val type = when {
--- a/client-android/serenada-call-ui/src/main/java/app/serenada/callui/CallScreen.kt
+++ b/client-android/serenada-call-ui/src/main/java/app/serenada/callui/CallScreen.kt
@@ -309,12 +309,11 @@ internal fun CallScreen(
       androidx.compose.runtime.CompositionLocalProvider(LocalAvatarCache provides avatarCache) {
         if (isSystemPictureInPicture) {
             val feed =
-                when {
-                    effectiveLocalLarge && uiState.localVideoEnabled -> SystemPictureInPictureFeed.Local
-                    uiState.remoteParticipants.firstOrNull() != null -> SystemPictureInPictureFeed.Remote
-                    uiState.localVideoEnabled -> SystemPictureInPictureFeed.Local
-                    else -> SystemPictureInPictureFeed.Remote
-                }
+                selectSystemPictureInPictureFeed(
+                    localIsLarge = effectiveLocalLarge,
+                    localVideoEnabled = uiState.localVideoEnabled,
+                    hasRemote = uiState.remoteParticipants.isNotEmpty(),
+                )
             SystemPictureInPictureContent(
                 uiState = uiState,
                 feed = feed,
--- a/client-android/serenada-call-ui/src/main/java/app/serenada/callui/CallScreen.kt
+++ b/client-android/serenada-call-ui/src/main/java/app/serenada/callui/CallScreen.kt
@@ -119,6 +119,7 @@ internal fun CallScreen(
     config: SerenadaCallFlowConfig = SerenadaCallFlowConfig(),
     theme: SerenadaCallFlowTheme = SerenadaCallFlowTheme(),
     strings: Map<SerenadaString, String>? = null,
+    isSystemPictureInPicture: Boolean = false,
     onToggleAudio: () -> Unit,
     onToggleVideo: () -> Unit,
     onFlipCamera: () -> Unit,
@@ -259,8 +260,13 @@ internal fun CallScreen(
             )
         }
 
+    val isCallChromeVisible = !isSystemPictureInPicture
+    val areControlsActuallyVisible = areControlsVisible && isCallChromeVisible
+
     val toggleControlsVisibility: () -> Unit = {
-        if (areControlsVisible) {
+        if (isSystemPictureInPicture) {
+            Unit
+        } else if (areControlsVisible) {
             areControlsVisible = false
             wereControlsLastHiddenByAutoHide = false
         } else {
@@ -274,7 +280,7 @@ internal fun CallScreen(
 
     // Auto-hide controls
     LaunchedEffect(areControlsVisible, uiState.phase, isControlsAutoHideEnabled) {
-        if (areControlsVisible && uiState.phase == CallPhase.InCall && isControlsAutoHideEnabled) {
+        if (areControlsActuallyVisible && uiState.phase == CallPhase.InCall && isControlsAutoHideEnabled) {
             delay(8000)
             wereControlsLastHiddenByAutoHide = true
             areControlsVisible = false
@@ -294,12 +300,33 @@ internal fun CallScreen(
     // remote-as-large so the user doesn't see a giant "Camera off" placeholder.
     // The user's swap preference (`isLocalLarge`) is preserved and reapplied
     // automatically when video comes back on.
-    val effectiveLocalLarge = isLocalLarge && uiState.localVideoEnabled
+    val effectiveLocalLarge =
+        isLocalLarge && uiState.localVideoEnabled
 
     val avatarCache = rememberAvatarCache(config.avatarProvider)
 
     SerenadaTheme(theme) {
       androidx.compose.runtime.CompositionLocalProvider(LocalAvatarCache provides avatarCache) {
+        if (isSystemPictureInPicture) {
+            val feed =
+                when {
+                    effectiveLocalLarge && uiState.localVideoEnabled -> SystemPictureInPictureFeed.Local
+                    uiState.remoteParticipants.firstOrNull() != null -> SystemPictureInPictureFeed.Remote
+                    uiState.localVideoEnabled -> SystemPictureInPictureFeed.Local
+                    else -> SystemPictureInPictureFeed.Remote
+                }
+            SystemPictureInPictureContent(
+                uiState = uiState,
+                feed = feed,
+                eglContext = eglContext,
+                localContentScale = if (uiState.isScreenSharing) ContentScale.Fit else ContentScale.Crop,
+                remoteContentScale = if (remoteVideoFitCover) ContentScale.Crop else ContentScale.Fit,
+                attachLocalSink = attachLocalSink,
+                detachLocalSink = detachLocalSink,
+                attachRemoteSink = attachRemoteSink,
+                detachRemoteSink = detachRemoteSink,
+            )
+        } else {
         BoxWithConstraints(
             modifier =
                 Modifier.fillMaxSize().background(theme.backgroundColor)
@@ -327,7 +354,7 @@ internal fun CallScreen(
                         uiState.connectionState == "CONNECTED")
         val animatedPipBottomPadding by
         animateDpAsState(
-            targetValue = if (areControlsVisible) 130.dp else 48.dp,
+            targetValue = if (areControlsActuallyVisible) 130.dp else 48.dp,
             animationSpec = tween(durationMillis = controlsAnimationDuration),
             label = "pip_bottom_padding"
         )
@@ -340,7 +367,7 @@ internal fun CallScreen(
         val pipInnerCornerRadius =
             if (pipCornerRadius > pipContentPadding) pipCornerRadius - pipContentPadding else 0.dp
         val mainVideoBottomPadding by animateDpAsState(
-            targetValue = if (areControlsVisible) 134.dp else 0.dp,
+            targetValue = if (areControlsActuallyVisible) 134.dp else 0.dp,
             animationSpec = tween(durationMillis = controlsAnimationDuration),
             label = "main_video_bottom_padding"
         )
@@ -611,7 +638,7 @@ internal fun CallScreen(
         }
 
         // Debug tap target — only when debug overlay is enabled via config
-        if (config.debugOverlayEnabled) {
+        if (config.debugOverlayEnabled && isCallChromeVisible) {
             Box(
                 modifier =
                     Modifier.align(Alignment.TopStart)
@@ -648,7 +675,7 @@ internal fun CallScreen(
         }
 
         // Waiting State Overlay
-        if (uiState.phase == CallPhase.Waiting && !effectiveLocalLarge) {
+        if (uiState.phase == CallPhase.Waiting && !effectiveLocalLarge && isCallChromeVisible) {
             WaitingOverlay(
                 roomShareUrl = roomShareUrl,
                 onInviteToRoom = onInviteToRoom,
@@ -660,7 +687,7 @@ internal fun CallScreen(
 
         // Reconnecting Indicator
         AnimatedVisibility(
-            visible = showReconnectingBadge,
+            visible = showReconnectingBadge && isCallChromeVisible,
             enter = fadeIn(),
             exit = fadeOut(),
             modifier = Modifier.align(Alignment.TopCenter).padding(top = 64.dp)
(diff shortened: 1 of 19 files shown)


agatx:
@codex please review

client-android/serenada-call-ui/src/main/java/app/serenada/callui/SystemPictureInPicture.kt
```diff
@@ -0,0 +1,288 @@
...
+        }
+        componentActivity.addOnUserLeaveHintListener(listener)
+        onDispose {
+            componentActivity.removeOnUserLeaveHintListener(listener)
+        }
+    }
+
+    LaunchedEffect(activity, supported, isInPictureInPicture, shouldEnter) {
+        if (supported && isInPictureInPicture && !shouldEnter) {
+            runCatching { activity.finish() }
```
agatx: **[Correctness / High] SDK finishes the host Activity on PiP exit**

**The SDK calls `activity.finish()` on the host's Activity when a call ends while in system PiP.**

When the call leaves a PiP-eligible phase while in PiP, this `LaunchedEffect` runs `activity.finish()` (and the `onDispose` at line 140 does the same whenever the composable leaves composition while still in PiP). `findActivity()` walks up to the host's Activity, so in a single-Activity host — the documented integration pattern, and exactly what the bundled sample's `MainActivity` is — this tears down the whole host Activity and its back stack instead of dismissing only the call screen.

**Scenario:** user backgrounds an active call into PiP, remote leaves → phase becomes idle → `shouldEnter` flips false → `finish()` kills `MainActivity`. The app exits / loses unrelated screens instead of returning to the join/home screen.

The SDK already has an `onDismiss` contract (SerenadaCallFlow wires it to `navController.popBackStack()` and the normal end path routes through it). Recommend signaling the host via `onDismiss` (or a dedicated `onPipDismissRequested`) instead of reaching around it to `finish()` the Activity directly.

client-android/serenada-call-ui/src/main/java/app/serenada/callui/SystemPictureInPicture.kt
```diff
@@ -0,0 +1,288 @@
...
+                    mirror = uiState.isFrontCamera && !uiState.isScreenSharing,
+                    contentScale = localContentScale,
+                )
+            }
+            showRemoteVideo -> {
+                TextureVideoSurface(
+                    modifier = Modifier.fillMaxSize(),
+                    rendererName = "system-pip-remote-${remote.cid}",
+                    eglContext = eglContext,
+                    onAttach = attachRemoteSink,
```
agatx: **[Correctness / Med-High] Multi-party PiP shows the wrong remote**

**Android system PiP always renders the first remote, ignoring the spotlighted/pinned participant.**

The remote feed is attached via the cid-less `attachRemoteSink`, and `selectSystemPictureInPictureFeed` returns only `Local`/`Remote` with no participant id. `attachRemoteSink` resolves to the first non-suspended remote in `SerenadaSession`. In a multi-party Frontline call (`maxParticipants = 4` on this branch, with full spotlight/pin plumbing) where the user spotlights remote C and backgrounds to PiP, the window shows remote A instead of C.

iOS threads the pinned/spotlight cid into `.remote(cid:)` and calls `attachRemoteRenderer(_:forCid:)`, so the two platforms diverge and Android shows the wrong person. (1:1 calls are unaffected, since `first` == the only remote.)

Recommend extending `selectSystemPictureInPictureFeed`/`SystemPictureInPictureContent` to carry a cid and using `attachRemoteSinkForCid`, matching the on-screen spotlight and iOS.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
```diff
@@ -0,0 +1,889 @@
...
+    func updateUIView(_ uiView: SystemPictureInPictureSourceView, context: Context) {
+        context.coordinator.rendererProvider = rendererProvider
+        context.coordinator.update(
+            enabled: enabled,
+            participant: participant,
+            avatarImage: avatarImage
+        )
+    }
+
+    static func dismantleUIView(_ uiView: UIView, coordinator: Coordinator) {
```
agatx: **[Correctness / Med-High] dismantleUIView leaves system PiP running and leaks the controller**

**On teardown, PiP is never stopped and the controller + renderer leak.**

`dismantleUIView` only calls `coordinator.detachRenderer()`. It never calls `stopPictureInPicture()` nor releases `controller`/`contentController`, and `Coordinator` has no `deinit`. The only `stopPictureInPicture()` is inside `update(enabled:false)`.

When the host removes the whole `SerenadaCallFlow` subtree while PiP is active — e.g. `endCall()` → `onDismiss?()`, or a navigation pop; with `canStartPictureInPictureAutomaticallyFromInline = true` PiP can be active after the app was backgrounded — SwiftUI tears down the overlay and calls `dismantleUIView` directly, with no guaranteed interleaved `enabled=false` update first. Result: the system PiP window stays on screen showing a frozen last frame, and the `AVPictureInPictureController` plus its `RTCVideoRenderer` video view leak for the process lifetime.

Recommend calling `stopPictureInPicture()` and niling `controller`/`contentController` in `dismantleUIView`, and/or adding a `Coordinator.deinit`.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
```diff
@@ -0,0 +1,889 @@
...
+    }
+
+    private var selectedParticipant: SystemPictureInPictureParticipant {
+        switch source {
+        case .local:
+            return SystemPictureInPictureParticipant(
+                source: .local,
+                videoEnabled: uiState.localVideoEnabled,
+                displayName: uiState.localDisplayName,
+                peerId: uiState.localCid,
```
agatx, line 365: **[Correctness / Med] Local-feed PiP avatar never resolves**

**The local user's configured avatar can't load in PiP; it always falls back to initials.**

For the `.local` feed, `selectedParticipant` sets `peerId: uiState.localCid`, and `SystemPictureInPictureLayer.body` calls `avatarCache?.image(for: localCid)` / `.task { avatarCache?.load(peerId: localCid) }`. But `localCid` is the per-call, server-issued connection id — explicitly distinct from the host-supplied `peerId` that `AvatarProvider` resolves — and the cache is only ever populated with remote `peerId`s. So when PiP shows the local feed with the camera off, the configured local avatar never resolves and PiP shows generic initials/"•".

Degraded display, not a crash. `CallUiState` carries no local `appPeerId`; resolving the local avatar would require threading the host's local peer identity through (the normal-UI local placeholder, `FrontlineLocalAvatar`, sidesteps this by being initials-only).

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
```diff
@@ -0,0 +1,889 @@
...
+                    provider?.attachLocalRenderer(videoView)
+                case .remote(let cid):
+                    if let cid {
+                        provider?.attachRemoteRenderer(videoView, forCid: cid)
+                    } else {
+                        provider?.attachRemoteRenderer(videoView)
+                    }
+                }
+            }
+            attachedSource = source
```
agatx, line 538: **[Correctness / Med — plausible] PiP renderer attach/detach bookkeeping can desync**

**`attachedSource` is updated synchronously but the real attach/detach is deferred, so bookkeeping can lead reality.**

`attachedSource` is set synchronously here while the actual `attach/detachLocalRenderer` / `attachRemoteRenderer` run inside `Task { @MainActor }` (deferred, because `CallRendererProvider` is `@MainActor`-isolated and `Coordinator` is not). `detachRenderer` computes its detach target synchronously from this optimistic `attachedSource`, and the early `guard let videoView … else { return }` (≈line 468) can return without updating `attachedSource`, leaving it desynced from the actually-attached track.

Under rapid source changes (multi-party pin flipping) or the brief nil-`videoView` startup window, the single shared `videoView` can end up attached to the wrong track, or detached when it should be showing (black PiP). Within a single `update()` the two queued Tasks are FIFO-correct, so this needs tight timing — hence plausible rather than guaranteed.

Consider making `Coordinator` `@MainActor` (so attach/detach run synchronously) or updating `attachedSource` inside the same hop as the provider call.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
```diff
@@ -0,0 +1,889 @@
...
+    @Environment(\.avatarCache) private var avatarCache
+
+    var body: some View {
+        let participant = selectedParticipant
+        let avatarImage = participant.peerId.flatMap { avatarCache?.image(for: $0) }
+
+        GeometryReader { geometry in
+            let frame = normalizedSourceFrame(in: geometry.size)
+            SystemPictureInPictureHost(
+                enabled: enabled && uiState.phase.isSystemPictureInPicturePhase,
```
agatx, line 329: **[Efficiency / Med] PiP renderer processes every frame during normal foreground calls**

**The off-screen PiP renderer runs for the entire call, not just while PiP is visible.**

`enabled` here is `enabled && uiState.phase.isSystemPictureInPicturePhase` — true throughout waiting/inCall whenever `systemPictureInPictureEnabled` is set (both sample hosts set it), independent of whether system PiP is actually active. `Coordinator.update(enabled:true)` then unconditionally calls `attachRenderer`, so `SystemPictureInPictureVideoView` is a live `RTCVideoRenderer` for the whole call.

Per frame it does a `CVPixelBufferCreate` + `CIContext.render` (rotated frames), a software I420→NV12 interleave (non-`RTCCVPixelBuffer` frames), and a `CMSampleBuffer` creation — all on top of the on-screen renderer, during every ordinary foreground video call. That's continuous, avoidable CPU/GPU/battery cost.

Recommend gating the attach on `controller.isPictureInPictureActive` and attaching lazily in `pictureInPictureControllerWillStartPictureInPicture` (detaching on stop).

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
```diff
@@ -0,0 +1,889 @@
...
+    let width = Int(i420.width)
+    let height = Int(i420.height)
+    guard width > 0, height > 0 else { return nil }
+    var buffer: CVPixelBuffer?
+    let attrs: [String: Any] = [kCVPixelBufferIOSurfacePropertiesKey as String: [:]]
+    let status = CVPixelBufferCreate(
+        kCFAllocatorDefault,
+        width,
+        height,
+        kCVPixelFormatType_420YpCbCr8BiPlanarFullRange,
```
agatx, line 897: **[Correctness / Low — plausible] I420→CVPixelBuffer labels video-range data as full-range**

**Video-range luma/chroma is copied verbatim into a `…FullRange` buffer, slightly washing out PiP video.**

The destination is `kCVPixelFormatType_420YpCbCr8BiPlanarFullRange` and the I420 bytes are copied as-is with no color attachments. WebRTC frames are typically video-range (16–235); interpreting them as full-range stretches levels (lighter / higher-contrast) for the software path.

Low severity and not a new regression: this is byte-identical to the already-shipped `FrameSnapshotCapturer.i420ToCVPixelBuffer` (v0.6.12), so the delta is apparently tolerated. Flagging for awareness — de-duping against the SerenadaCore helper (see the reuse comment) would let a single fix address both.

client-ios/SerenadaCore/Sources/Call/CameraCaptureController.swift
```diff
@@ -359,6 +359,7 @@ final class CameraCaptureController {
 
         let fps = selectCaptureFPS(for: format)
 
+        enableMultitaskingCameraAccessIfSupported(on: capturer.captureSession)
```
agatx, line 362: **[Altitude / Low] Multitasking camera access is enabled in core for ALL consumers, ungated by the PiP flag**

**A UI-layer opt-in silently mutates core capture behavior for every SDK consumer.**

`enableMultitaskingCameraAccessIfSupported` runs unconditionally on every capture start (here and in `CompositeCameraVideoCapturer`), guarded only by `isMultitaskingCameraAccessSupported`. The opt-in flag `systemPictureInPictureEnabled` lives in the UI layer (`SerenadaCallFlowConfig`) and never reaches SerenadaCore (the dependency only goes UI→core), so `isMultitaskingCameraAccessEnabled = true` is set for every consumer regardless of whether they enabled PiP.

Impact today is benign (it's effectively a no-op unless the app actually enters a multitasking/PiP state), so this is an altitude/layering note rather than a bug. Consider threading the PiP flag into core so the enablement matches the opt-in instead of applying to all consumers.

client-ios/SerenadaCallUI/Sources/CallScreen.swift
```diff
@@ -699,6 +716,32 @@ struct CallScreenView: View {
         return nil
     }
 
+    private func currentSystemPictureInPictureSource(
```
agatx: **[Altitude / Med] PiP feed-selection logic is duplicated across both iOS call screens**

**Three independent feed-selection ladders, no shared selector, already diverging.**

`currentSystemPictureInPictureSource` here and the near-identical one in `FrontlineCallScreen.swift:233` both encode the same 1:1 priority order (local-primary → first remote → local-if-video → `remote(nil)`), differing only in the multi-party head (`pinnedParticipantId` vs `pinnedSpotlightId/selectedSpotlightId/lastVideoStartedParticipantId`).

Android deliberately centralizes this in one `selectSystemPictureInPictureFeed` "so the priority order stays in one place." With three ladders (1 Android + 2 iOS) and no parity check (contrast `scripts/check-resilience-constants.mjs`), they will drift — and they already diverge from Android, which gates the local-large branch on `localVideoEnabled`. This duplication is the root cause of the multi-party mismatch flagged on the Android side.

Recommend one shared iOS selector parameterized by each variant's spotlight inputs.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
```diff
@@ -0,0 +1,889 @@
...
+            sourceView?.update(
+                participant: currentParticipant,
+                avatarImage: currentAvatarImage,
+                placeholderVisible: true
+            )
+        }
+    }
+}
+
+private final class SystemPictureInPictureContentController: AVPictureInPictureVideoCallViewController {
```
agatx, line 591: **[Cleanup / Med] The PiP avatar/initials/name placeholder is built twice**

**Two near-identical UIKit placeholder hierarchies with diverging scaling formulas.**

`SystemPictureInPictureContentController` (here) and `SystemPictureInPictureSourceView` (above) each build a circular avatar `UIImageView` + initials `UILabel` + name `UILabel` (the source view also a live timer) with near-identical constraints, and two *different* scaling routines — `updateLayoutMetrics` uses `min(1, w/140, h/244)`, `updatePlaceholderLayoutMetrics` uses `width*0.36` clamped — which will drift.

The source view is the in-window `activeVideoCallSourceView`, covered by the live call UI while in-app and replaced by the content controller while in PiP, so its placeholder is rarely the surface the user actually sees. Recommend factoring one reusable placeholder (or a `UIHostingController` wrapping the existing `RemoteAvatarView`) for both — or dropping the source-view placeholder content entirely and keeping it a plain anchor.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
```diff
@@ -0,0 +1,889 @@
...
+            Self.pixelBufferAttributes,
+            &output
+        )
+        guard status == kCVReturnSuccess, let output else { return nil }
+        ciContext.render(image, to: output)
+        return output
+    }
+}
+
+private func i420ToCVPixelBuffer(_ i420: RTCI420BufferProtocol) -> CVPixelBuffer? {
```
agatx, line 887: **[Reuse / Med] i420ToCVPixelBuffer + cgOrientation duplicate existing SerenadaCore helpers**

**Near-verbatim copies of helpers already in SerenadaCore.**

`i420ToCVPixelBuffer` (here) and `cgOrientation(for:)` (line 880) duplicate the helpers in `client-ios/SerenadaCore/Sources/Call/FrameSnapshotCapturer.swift` — `cgOrientation` is byte-identical; `i420ToCVPixelBuffer` differs only by a dropped explanatory comment. Two copies of the NV12 conversion + rotation mapping will drift (and share the full-range color note above).

Caveat: the SerenadaCore versions are top-level `private func`, so this isn't a free "call the existing function" — de-duping means promoting them to `public`/`package` in SerenadaCore (keeping the `#if canImport(WebRTC)` guard) and deleting both UI-module copies. Worth doing to keep a single conversion path.

client-android/serenada-call-ui/src/main/java/app/serenada/callui/CallScreen.kt
```diff
@@ -259,8 +260,13 @@ internal fun CallScreen(
             )
         }
 
+    val isCallChromeVisible = !isSystemPictureInPicture
```
agatx: **[Cleanup / Med] PiP chrome guards are dead code in the standard CallScreen**

**`isCallChromeVisible` / `areControlsActuallyVisible` are constant in CallScreen's structure.**

The chrome that reads these lives inside the `else` branch of `if (isSystemPictureInPicture) { SystemPictureInPictureContent(...) } else { BoxWithConstraints {…} }`, where `isSystemPictureInPicture` is constant-false. So `isCallChromeVisible` is always `true` and `areControlsActuallyVisible == areControlsVisible` here. The added `&& isCallChromeVisible` guards (debug tap, waiting overlay, reconnecting badge, snapshot/flash/fit) and the `if (isSystemPictureInPicture) { Unit }` branch in `toggleControlsVisibility` are therefore redundant in this file.

(They ARE load-bearing in `FrontlineCallScreen`, which renders PiP content in the *same* subtree as chrome — so keep them there.) Recommend reverting the CallScreen guards to plain `areControlsVisible` and dropping the unused flags. Minor: `effectiveLocalLarge` was also reflowed onto two lines with no logic change.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
```diff
@@ -0,0 +1,889 @@
...
+            if let frame, !frame.isEmpty {
+                onFrameChanged(frame)
+            } else {
+                onFrameChanged(nil)
+            }
+        }
+    }
+}
+
+private struct SystemPictureInPictureSourceFramePreferenceKey: PreferenceKey {
```
agatx, line 78: **[Cleanup / Low — plausible] Per-surface frame reporting via PreferenceKey is fragile and likely unnecessary**

**A custom PreferenceKey hand-attached to ~6 surfaces, with a stale-keeping reduce.**

The source frame is reported by attaching `.systemPictureInPictureSourceFrame(onChange:)` to ~6 primary surfaces across CallScreen/FrontlineCallScreen (and pushed into `FrontlineAudioLarge` as an extra param), feeding this `PreferenceKey` whose `reduce` (`value = nextValue() ?? value`) keeps a stale frame when two reporters briefly coexist during a transition. Any new primary-surface branch that forgets the modifier reports a nil/stale frame, so the PiP enter/exit transition flies from the wrong rect.

Since `SystemPictureInPictureLayer` already has a full-screen `GeometryReader` fallback and the host view is invisible (`allowsHitTesting(false)`), consider dropping the per-surface reporting and using the full-screen frame, or measuring a single "large surface" container once.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
```diff
@@ -0,0 +1,889 @@
...
+            timer = nil
+        } else if timer == nil {
+            timer = Timer.scheduledTimer(withTimeInterval: 1, repeats: true) { [weak self] _ in
+                self?.updateTimer()
+            }
+        }
+        updateTimer()
+    }
+
+    private func updateTimer() {
```
agatx, line 299: **[Reuse / Low] PiP timer duplicates FrontlineTimerLabel's elapsed formatting**

**`updateTimer()` reimplements the Frontline timer.**

This duplicates `FrontlineTimerLabel` (`FrontlineCallScreen.swift:960`): same `String(format: "%02lld:%02lld", elapsed/60, elapsed%60)`, same `max(0, (nowMs - startedAt)/1000)`, same fallback-start pattern — same module. Extract one shared `formatCallElapsed(startedAtMs:fallbackMs:)` so the two stay consistent. (Tied to the "placeholder built twice" note — this timer only exists on the rarely-seen source-view placeholder.)

client-android/serenada-call-ui/src/main/java/app/serenada/callui/SystemPictureInPicture.kt
```diff
@@ -0,0 +1,288 @@
...
+    autoEnter: Boolean,
+): PictureInPictureParams {
+    val builder =
+        PictureInPictureParams.Builder()
+            .setAspectRatio(sourceRect.toPipAspectRatio())
+    if (sourceRect != null && !sourceRect.isEmpty) {
+        builder.setSourceRectHint(sourceRect)
+    }
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+        builder.setAutoEnterEnabled(autoEnter)
```
agatx, line 158: **[Correctness / Low — plausible] PiP entry is double-driven on Android 12+**

**Both auto-enter and a manual enter fire on S+.**

On Android S+ this sets `setAutoEnterEnabled(autoEnter)`, while the `addOnUserLeaveHintListener { enterPictureInPictureMode(params) }` (line 122) is also active. On Home/gesture-leave the system auto-enter and the explicit `enterPictureInPictureMode` both fire; the redundant explicit call races the system transition and is only swallowed by `runCatching`, relying on the platform to no-op an in-progress enter.

Recommend gating the manual `onUserLeaveHint` enter to pre-S (`Build.VERSION.SDK_INT < S`), since `setAutoEnterEnabled` already covers S+.

client-android/serenada-call-ui/src/main/java/app/serenada/callui/SystemPictureInPicture.kt
```diff
@@ -0,0 +1,288 @@
...
+            componentActivity.removeOnPictureInPictureModeChangedListener(listener)
+        }
+    }
+
+    LaunchedEffect(activity, supported, shouldEnter, sourceRect) {
+        if (!supported) {
+            latestParams.value = null
+            return@LaunchedEffect
+        }
+        val hostActivity = activity
```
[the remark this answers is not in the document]
agatx: Fixed in 1733bc7c. The host Activity is now bound with an explicit non-null early return before applying PiP params, and Android assembleDebug passes.

client-android/serenada-call-ui/src/main/java/app/serenada/callui/SystemPictureInPicture.kt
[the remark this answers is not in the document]
agatx: Fixed in 1733bc7c. The direct Activity finish path was removed; PiP dismissal now routes through the call-flow dismiss contract.

client-android/serenada-call-ui/src/main/java/app/serenada/callui/SystemPictureInPicture.kt
```diff
@@ -0,0 +1,288 @@
...
+    LaunchedEffect(activity, supported, isInPictureInPicture, shouldEnter) {
+        if (supported && isInPictureInPicture && !shouldEnter) {
+            runCatching { activity.finish() }
+        }
+    }
+
+    DisposableEffect(activity, supported) {
+        onDispose {
+            if (supported && latestIsInPictureInPicture) {
+                runCatching { activity.finish() }
```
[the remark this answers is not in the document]
agatx: Fixed in 1733bc7c. The disposal-time Activity finish path was removed; teardown no longer reaches around the host Activity.

client-android/serenada-call-ui/src/main/java/app/serenada/callui/SystemPictureInPicture.kt
```diff
@@ -0,0 +1,288 @@
...
+    localContentScale: ContentScale,
+    remoteContentScale: ContentScale,
+    attachLocalSink: (VideoSink) -> Unit,
+    detachLocalSink: (VideoSink) -> Unit,
+    attachRemoteSink: (VideoSink) -> Unit,
+    detachRemoteSink: (VideoSink) -> Unit,
+) {
+    val remote = uiState.remoteParticipants.firstOrNull()
+    val showLocalVideo = feed == SystemPictureInPictureFeed.Local && uiState.localVideoEnabled
+    val showRemoteVideo = feed == SystemPictureInPictureFeed.Remote && remote?.videoEnabled == true
```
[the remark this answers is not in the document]
agatx: Fixed in 1733bc7c. Android PiP now carries the selected remote CID and uses the CID-specific sink attach/detach path.

client-android/serenada-call-ui/src/main/java/app/serenada/callui/SystemPictureInPicture.kt
```diff
@@ -0,0 +1,288 @@
...
+    feed: SystemPictureInPictureFeed,
+    eglContext: EglBase.Context,
+    localContentScale: ContentScale,
+    remoteContentScale: ContentScale,
+    attachLocalSink: (VideoSink) -> Unit,
+    detachLocalSink: (VideoSink) -> Unit,
+    attachRemoteSink: (VideoSink) -> Unit,
+    detachRemoteSink: (VideoSink) -> Unit,
+) {
+    val remote = uiState.remoteParticipants.firstOrNull()
```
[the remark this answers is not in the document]
agatx: Fixed in 1733bc7c. PiP selection now preserves the pinned/spotlighted remote CID instead of collapsing to the first remote.

client-android/serenada-call-ui/src/main/java/app/serenada/callui/SystemPictureInPicture.kt
agatx: Fixed in 1733bc7c. The SDK no longer calls Activity.finish(); when a call leaves a PiP-eligible phase while in PiP, it invokes the existing onDismiss contract.

client-android/serenada-call-ui/src/main/java/app/serenada/callui/SystemPictureInPicture.kt
agatx: Fixed in 1733bc7c. CallScreen and Frontline now resolve the selected remote CID and SystemPictureInPictureContent attaches via attachRemoteSinkForCid/detachRemoteSinkForCid.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
agatx: Fixed in 1733bc7c. dismantleUIView now stops active PiP, hides the placeholder, detaches the renderer, clears the delegate, and releases controller/content state.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
agatx: Fixed in 1733bc7c. The coordinator is now main-actor-bound and attach/detach bookkeeping is updated synchronously with the renderer provider calls.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
agatx: Fixed in 1733bc7c. The PiP renderer now attaches only while PiP is active or starting, and detaches again on stop.

client-android/serenada-call-ui/src/main/java/app/serenada/callui/CallScreen.kt
agatx: Fixed in 1733bc7c. The standard CallScreen PiP subtree is separate, so I removed the redundant chrome guards and restored normal control visibility in the non-PiP subtree.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
agatx: Fixed in 1733bc7c. Added a shared elapsed-call formatter used by both the Frontline timer and PiP source placeholder.

client-android/serenada-call-ui/src/main/java/app/serenada/callui/SystemPictureInPicture.kt
agatx: Fixed in 1733bc7c. Manual OnUserLeaveHint PiP entry is now registered only below Android S; S+ relies on auto-enter.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
agatx: Agreed this is a real limitation, but I am not threading a new local app peer ID through CallUiState in this PR. Current in-call local placeholders are initials-only too; remote avatars still resolve through peerId. Leaving this as a follow-up API decision.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
agatx: Agreed and not fixing in this PR. This matches the already-shipped snapshot conversion helper, so changing range semantics should be done once in a shared conversion path rather than only for PiP.

client-ios/SerenadaCore/Sources/Call/CameraCaptureController.swift
agatx: Agreed as a layering concern. I am leaving this out of this PR because plumbing the UI opt-in into core broadens the core contract; current behavior is support-gated and benign unless the app enters multitasking/PiP.

client-ios/SerenadaCallUI/Sources/CallScreen.swift
agatx: Agreed as cleanup. Android had the correctness bug and is fixed in 1733bc7c; iOS already threads the selected CID. I am not refactoring the iOS selector plumbing in this PR.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
agatx: Agreed as cleanup. I kept the source-view placeholder because it is part of the restore anchor/transition behavior we were tuning; factoring or dropping it is better handled after transition QA settles.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
agatx: Agreed. I am not promoting the private SerenadaCore helpers to package/public API in this PR; that is a broader reuse/API change and should be paired with the range-semantics fix above.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
agatx: Agreed this is fragile, but the per-surface rect is intentional for the smoother PiP transition path. I am keeping it for this PR and would revisit only if a coarser anchor preserves transition quality.

client-ios/SerenadaCallUI/Sources/CallScreen.swift
agatx: Fixed in bdf9c576. iOS now uses a shared selectSystemPictureInPictureSource helper for the standard and Frontline call screens; each screen only passes its local/remote spotlight candidates, including Frontline content-id normalization. Added CallScreenStateTests coverage for the shared priority order and preferred source handling.

PR #124, opened 2026-06-05
Fix web audio routing and bump SDK to 0.8.4

agatx:
## Summary
- Fix web local audio capture to prefer the browser default audio input before first capture, with default-output fallback on device changes.
- Refresh the local audio track after immediate and settled `devicechange` events without renegotiating already-active peer senders when replacement is enough.
- Share concurrent `startLocalMedia()` calls and bump the SDK to `0.8.4` across package metadata, platform constants, docs, and web sample lockfiles.

## Root Cause
Chrome can keep an existing built-in microphone track alive even after the browser default audio route points at a Bluetooth headset. The web SDK only captured generic audio constraints and only watched device changes for camera availability, so loopback tabs could send built-in mic audio while output correctly moved to Bluetooth.

## Checks
- `node scripts/check-version-parity.mjs`
- `npx eslint src/media/MediaEngine.ts` from `client/packages/core`
- `npm test -- test/media/MediaEngine.test.ts` from `client/packages/core` (409 tests)
- `npm run build` from `client` (passes with existing Vite large-chunk warning)
- `npm run build` from `samples/web`

### DIFF
--- a/client/packages/core/src/media/MediaEngine.ts
+++ b/client/packages/core/src/media/MediaEngine.ts
@@ -24,6 +24,7 @@ const ICE_STATE_PRIORITY: RTCIceConnectionState[] = ['failed', 'disconnected', '
 const CONN_STATE_PRIORITY: RTCPeerConnectionState[] = ['failed', 'disconnected', 'connecting', 'new', 'connected', 'closed'];
 const SIG_STATE_PRIORITY: RTCSignalingState[] = ['closed', 'have-local-offer', 'have-remote-offer', 'have-local-pranswer', 'have-remote-pranswer', 'stable'];
 const LEGACY_OFFER_ID = '__legacy__';
+const DEVICE_CHANGE_SETTLE_DELAY_MS = 2000;
 
 function getSignalingState(pc: RTCPeerConnection): RTCSignalingState {
     return pc.signalingState;
@@ -67,6 +68,12 @@ interface PeerState {
     lastOutboundMediaRecoveryAt: number;
 }
 
+interface PreferredAudioInputSelection {
+    device: MediaDeviceInfo | null;
+    currentMatchesPreferredRoute: boolean;
+    basis: 'default-input' | 'default-output' | 'already-matched' | 'none' | 'no-devices';
+}
+
 export interface MediaEngineConfig {
     turnsOnly?: boolean;
     logger?: SerenadaLogger;
@@ -104,6 +111,8 @@ export class MediaEngine {
     private requestingMedia = false;
     private destroyed = false;
     private cameraRecoveryInFlight = false;
+    private audioRecoveryInFlight = false;
+    private localMediaStartPromise: Promise<MediaStream | null> | null = null;
     private mediaRequestId = 0;
     private retryingTimer: number | null = null;
     private localVideoHeartbeatAt = Date.now();
@@ -117,6 +126,7 @@ export class MediaEngine {
     private onlineHandler: (() => void) | null = null;
     private networkChangeHandler: (() => void) | null = null;
     private deviceChangeHandler: (() => void) | null = null;
+    private deviceChangeSettleTimer: number | null = null;
     private turnsOnly: boolean;
     private logger?: SerenadaLogger;
     private offerSequence = 0;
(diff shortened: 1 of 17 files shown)


agatx:
@codex please review

client/packages/core/src/media/MediaEngine.ts
```diff
@@ -1358,12 +1514,115 @@ export class MediaEngine {
...
+            return {
+                device: defaultInput.groupId ? inputForGroup(defaultInput.groupId) ?? defaultInput : defaultInput,
+                currentMatchesPreferredRoute: false,
+                basis: 'default-input',
+            };
+        }
+
+        if (defaultOutput?.groupId && currentGroupId !== defaultOutput.groupId) {
+            return {
+                device: inputForGroup(defaultOutput.groupId),
```
agatx: **Needless mic re-acquisition when the default output device has no microphone in its group.**

When the system default audio *output* is a mic-less device (Bluetooth A2DP speaker, HDMI/monitor, AirPlay), `defaultOutput.groupId` has no matching `audioinput`, so `inputForGroup(defaultOutput.groupId)` returns `null`. This branch then returns `{ device: null, currentMatchesPreferredRoute: false, basis: 'default-output' }`.

`refreshLocalAudioTrack` (lines 1329-1335) only bails on basis `no-devices`/`none` or `currentMatchesPreferredRoute` — it does **not** bail on a null `device`. So it proceeds to `acquireAudioTrack(enabled, undefined)`, re-acquires the **same** default mic with a generic constraint, calls `replaceTrack` on every peer, and stops the old track. Because the route never "matches" (the output group has no input), this fires on `initial-route-check` and on **every** `devicechange` (twice — immediate + the 2s settled pass) for the whole call whenever a mic-less output device is the default. That's a common setup (external monitor speakers, BT speaker).

Fix: when `inputForGroup(defaultOutput.groupId)` is null, keep the current input (treat as already-matched) instead of returning a re-acquire-triggering state — or add `if (!preferredInput.device) return false;` to the guard in `refreshLocalAudioTrack`.

client/packages/core/src/media/MediaEngine.ts
```diff
@@ -288,11 +304,17 @@ export class MediaEngine {
 
             this.applySpeechTrackHints(stream);
             this.localStream = stream;
-            await this.detectCameras();
+            const postStartDevices = await this.enumerateMediaDevices();
+            await this.detectCameras(postStartDevices);
             this.requestingMedia = false;
+            if (this.roomState && this.clientId) {
+                await this.refreshLocalAudioTrack('initial-route-check', postStartDevices, false);
```
agatx: **Initial route check with `updatePeers=false` can leave a peer sending a stopped audio track (race).**

This call relies on the attach loop at line 316 to propagate the swapped audio track to peers. But `attachLocalTrackToPeer` (line 1408) returns early **without replacing** when the transceiver's sender already holds a same-kind track:

```ts
if (transceiver.sender.track?.kind === track.kind) {
    return this.needsLocalTrackNegotiation(transceiver);
}
```

If a peer is created with the pre-refresh audio track during one of the `await`s above (an incoming `offer`/`ice` processed between `this.localStream = stream` at line 306 and this attach loop — `getOrCreatePeer`/`applyRemoteOffer` add the then-current track), the refresh swaps `this.localStream`'s audio and **stops** the old track. With `updatePeers=false` that peer is skipped, and the attach loop won't replace its existing sender track — so it keeps sending the stopped/dead track. Race-dependent (needs a peer created in that window), but the outcome is a silent loss of outbound audio to that peer. Passing `updatePeers=true` here, or having the attach path replace a stale track, would close it.

client/packages/core/src/media/MediaEngine.ts
```diff
@@ -1358,12 +1514,115 @@ export class MediaEngine {
...
+        return constraints;
+    }
+
+    private getTrackGroupId(track: MediaStreamTrack | null): string | null {
         try {
-            const devices = await navigator.mediaDevices.enumerateDevices();
-            this.hasMultipleCameras = devices.filter(d => d.kind === 'videoinput').length > 1;
-        } catch { /* ignore */ }
+            const groupId = track?.getSettings().groupId;
+            return groupId ? groupId : null;
```
agatx, line 1554: **Empty-string `groupId` collapses to `null`, causing repeated re-acquisition where `groupId` isn't exposed.**

`getTrackGroupId` returns `null` for an empty-string `groupId`, and `selectPreferredAudioInput` treats `!defaultInput.groupId` as "no group". On browsers/permission states that don't populate `groupId` on tracks/devices, `currentGroupId` is always `null`, so the `default-input` branch (line 1578) matches on every `devicechange` and re-acquires `deviceId: 'default'`, replacing the live audio track needlessly each time. Lower severity (most browsers populate `groupId` after permission is granted), but worth guarding — e.g. skip the refresh when the group identity is unknown on both the current track and the default device.

PR #130, opened 2026-06-15
[codex] Bump Serenada SDK to 0.8.6

agatx:
## Summary
- Stop iOS system Picture in Picture when the app returns to foreground so the inline call UI is restored cleanly.
- Bump shared Serenada SDK version surfaces to `0.8.6` and refresh local lockfile metadata.
- Update PiP docs and enable system PiP in the iOS sample with the required background modes.

## Release Scope
- iOS-only fix.
- No `web-release-*` tag was created or pushed, so this PR does not publish web packages.

## Validation
- `node scripts/check-version-parity.mjs`
- `plutil -lint samples/ios/Resources/Info.plist`
- `cd samples/ios && xcodegen generate`
- `cd samples/ios && xcodebuild -project SerenadaiOSSample.xcodeproj -scheme SerenadaiOSSample -destination 'generic/platform=iOS Simulator' build`

Co-authored by Codex (GPT-5 medium)

### DIFF
--- a/client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
+++ b/client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
@@ -425,7 +425,7 @@ private struct SystemPictureInPictureHost: UIViewRepresentable {
         init(rendererProvider: CallRendererProvider) {
             self.rendererProvider = rendererProvider
             super.init()
-            registerForegroundObservers()
+            registerForegroundObserver()
         }
 
         deinit {
@@ -503,7 +503,7 @@ private struct SystemPictureInPictureHost: UIViewRepresentable {
             currentParticipant = nil
             currentAvatarImage = nil
             isStoppingForForegroundActivation = false
-            removeForegroundObservers()
+            removeForegroundObserver()
         }
 
         private func ensureController() {
@@ -579,9 +579,9 @@ private struct SystemPictureInPictureHost: UIViewRepresentable {
             isStoppingForForegroundActivation = false
         }
 
-        private func registerForegroundObservers() {
+        private func registerForegroundObserver() {
             foregroundObserver = NotificationCenter.default.addObserver(
-                forName: UIApplication.willEnterForegroundNotification,
+                forName: UIApplication.didBecomeActiveNotification,
                 object: nil,
                 queue: .main
             ) { [weak self] _ in
@@ -591,7 +591,7 @@ private struct SystemPictureInPictureHost: UIViewRepresentable {
             }
         }
 
-        private func removeForegroundObservers() {
+        private func removeForegroundObserver() {
             if let foregroundObserver {
                 NotificationCenter.default.removeObserver(foregroundObserver)
                 self.foregroundObserver = nil
--- a/client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
+++ b/client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
@@ -419,10 +419,13 @@ private struct SystemPictureInPictureHost: UIViewRepresentable {
         private var attachedSource: SystemPictureInPictureSource?
         private var currentParticipant: SystemPictureInPictureParticipant?
         private var currentAvatarImage: UIImage?
+        private var foregroundObservers: [NSObjectProtocol] = []
+        private var isStoppingForForegroundActivation = false
 
         init(rendererProvider: CallRendererProvider) {
             self.rendererProvider = rendererProvider
             super.init()
+            registerForegroundObservers()
         }
 
         func configure(sourceView: SystemPictureInPictureSourceView) {
@@ -493,6 +496,7 @@ private struct SystemPictureInPictureHost: UIViewRepresentable {
             contentController = nil
             currentParticipant = nil
             currentAvatarImage = nil
+            removeForegroundObservers()
         }
 
         private func ensureController() {
@@ -557,12 +561,39 @@ private struct SystemPictureInPictureHost: UIViewRepresentable {
         }
 
         func pictureInPictureControllerWillStopPictureInPicture(_ pictureInPictureController: AVPictureInPictureController) {
-            updateSourcePlaceholder(visible: true, animated: false)
+            updateSourcePlaceholder(visible: !isStoppingForForegroundActivation, animated: false)
         }
 
         func pictureInPictureControllerDidStopPictureInPicture(_ pictureInPictureController: AVPictureInPictureController) {
             updateSourcePlaceholder(visible: false, animated: true)
             detachRenderer()
+            isStoppingForForegroundActivation = false
+        }
+
+        private func registerForegroundObservers() {
+            let notifications = [
+                UIApplication.willEnterForegroundNotification,
+                UIApplication.didBecomeActiveNotification
+            ]
+            foregroundObservers = notifications.map { name in
+                NotificationCenter.default.addObserver(forName: name, object: nil, queue: .main) { [weak self] _ in
+                    Task { @MainActor [weak self] in
+                        self?.stopPictureInPictureForForegroundActivation()
+                    }
+                }
+            }
+        }
+
+        private func removeForegroundObservers() {
+            foregroundObservers.forEach(NotificationCenter.default.removeObserver)
+            foregroundObservers.removeAll()
+        }
+
+        private func stopPictureInPictureForForegroundActivation() {
+            guard controller?.isPictureInPictureActive == true else { return }
+            isStoppingForForegroundActivation = true
+            sourceView?.setPlaceholderVisible(false, animated: false)
+            controller?.stopPictureInPicture()
         }
 
         private func updateSourcePlaceholder(visible: Bool, animated: Bool) {
--- a/client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
+++ b/client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
@@ -419,13 +419,19 @@ private struct SystemPictureInPictureHost: UIViewRepresentable {
         private var attachedSource: SystemPictureInPictureSource?
         private var currentParticipant: SystemPictureInPictureParticipant?
         private var currentAvatarImage: UIImage?
-        private var foregroundObservers: [NSObjectProtocol] = []
+        private var foregroundObserver: NSObjectProtocol?
         private var isStoppingForForegroundActivation = false
 
         init(rendererProvider: CallRendererProvider) {
             self.rendererProvider = rendererProvider
             super.init()
             registerForegroundObservers()
+        }
+
+        deinit {
+            if let foregroundObserver {
+                NotificationCenter.default.removeObserver(foregroundObserver)
+            }
         }
 
         func configure(sourceView: SystemPictureInPictureSourceView) {
@@ -439,7 +445,7 @@ private struct SystemPictureInPictureHost: UIViewRepresentable {
             sourceView?.update(
                 participant: participant,
                 avatarImage: avatarImage,
-                placeholderVisible: controller?.isPictureInPictureActive == true
+                placeholderVisible: controller?.isPictureInPictureActive == true && !isStoppingForForegroundActivation
             )
 
             guard enabled else {
@@ -496,6 +502,7 @@ private struct SystemPictureInPictureHost: UIViewRepresentable {
             contentController = nil
             currentParticipant = nil
             currentAvatarImage = nil
+            isStoppingForForegroundActivation = false
             removeForegroundObservers()
         }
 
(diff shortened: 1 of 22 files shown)


client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
```diff
@@ -557,12 +561,39 @@ private struct SystemPictureInPictureHost: UIViewRepresentable {
...
+
+        private func removeForegroundObservers() {
+            foregroundObservers.forEach(NotificationCenter.default.removeObserver)
+            foregroundObservers.removeAll()
+        }
+
+        private func stopPictureInPictureForForegroundActivation() {
+            guard controller?.isPictureInPictureActive == true else { return }
+            isStoppingForForegroundActivation = true
+            sourceView?.setPlaceholderVisible(false, animated: false)
```
agatx, line 605: **Foreground stop can flicker the placeholder back on for a video-off participant.** This hides the source placeholder, but `update(enabled:)` (line 442) still derives `placeholderVisible: controller?.isPictureInPictureActive == true`. PiP stays "active" until `didStop` fires, so any `updateUIView` → `update(...)` during the in-flight stop re-shows the avatar/initials over the inline tile. The source view's `update` only gates on `!participant.videoEnabled`, so this lands on audio-only / video-off participants. The new `isStoppingForForegroundActivation` flag is never consulted in `update()`, so the SwiftUI render path fights the imperative hide and reintroduces the exact flash this PR targets. Have `update()` (or the source view) honor the stopping intent so the placeholder stays hidden until `didStop`.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
```diff
@@ -557,12 +561,39 @@ private struct SystemPictureInPictureHost: UIViewRepresentable {
...
 
         func pictureInPictureControllerWillStopPictureInPicture(_ pictureInPictureController: AVPictureInPictureController) {
-            updateSourcePlaceholder(visible: true, animated: false)
+            updateSourcePlaceholder(visible: !isStoppingForForegroundActivation, animated: false)
         }
 
         func pictureInPictureControllerDidStopPictureInPicture(_ pictureInPictureController: AVPictureInPictureController) {
             updateSourcePlaceholder(visible: false, animated: true)
             detachRenderer()
+            isStoppingForForegroundActivation = false
```
agatx, line 579: **`isStoppingForForegroundActivation` is cleared only here, so it can get stuck `true` and suppress the placeholder on a later, unrelated stop.** `willStop` reads `visible: !isStoppingForForegroundActivation`. The flag is set in `stopPictureInPictureForForegroundActivation()` but reset only in `didStop`. If `didStop` never arrives for that stop — e.g. `cleanup()` nils `controller`/`delegate` first (lines 494-495), or PiP auto-restarts via `canStartPictureInPictureAutomaticallyFromInline` before the stop completes (`willStart`/`didStart` don't reset it either) — the flag stays `true`. The next genuine manual PiP stop then evaluates `visible: false`, so the inline tile shows no avatar during the shrink-back, relocating the original artifact. Reset the flag in `cleanup()` and on the start callbacks (or scope it per-stop) so it can't leak across cycles.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
```diff
@@ -419,10 +419,13 @@ private struct SystemPictureInPictureHost: UIViewRepresentable {
         private var attachedSource: SystemPictureInPictureSource?
         private var currentParticipant: SystemPictureInPictureParticipant?
         private var currentAvatarImage: UIImage?
+        private var foregroundObservers: [NSObjectProtocol] = []
+        private var isStoppingForForegroundActivation = false
 
         init(rendererProvider: CallRendererProvider) {
             self.rendererProvider = rendererProvider
             super.init()
+            registerForegroundObservers()
```
agatx: **Block-based `NotificationCenter` observers have no `deinit` safety net.** They're added in `init` and removed only in `cleanup()`, which runs solely from `dismantleUIView`. Block observers registered via `addObserver(forName:object:queue:using:)` are not auto-removed on dealloc, so any teardown where SwiftUI doesn't deterministically call `dismantleUIView` leaks the observer tokens. The `[weak self]` capture keeps it from crashing, but the tokens and closures persist for the app's lifetime and fire a no-op on every foreground, accumulating one pair per call session. Add `deinit { removeForegroundObservers() }` as a belt-and-suspenders.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
```diff
@@ -557,12 +561,39 @@ private struct SystemPictureInPictureHost: UIViewRepresentable {
...
+        }
+
+        private func registerForegroundObservers() {
+            let notifications = [
+                UIApplication.willEnterForegroundNotification,
+                UIApplication.didBecomeActiveNotification
+            ]
+            foregroundObservers = notifications.map { name in
+                NotificationCenter.default.addObserver(forName: name, object: nil, queue: .main) { [weak self] _ in
+                    Task { @MainActor [weak self] in
```
agatx: **The `Task { @MainActor }` hop defers the stop by a runloop turn and widens the flash window.** The observer block is already delivered on `queue: .main`, so the wrapper exists only to satisfy `@MainActor` isolation. Making it a `Task` turns the call async, so SwiftUI's `updateUIView` can run in the gap and re-show the placeholder (see the race on `stopPictureInPictureForForegroundActivation`) before the stop actually executes. A synchronous main-actor dispatch (e.g. `MainActor.assumeIsolated { ... }` where the deployment target allows it) runs on the already-main thread and closes that window. The nested second `[weak self]` is also redundant since the outer `self` is already a weak optional.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
```diff
@@ -557,12 +561,39 @@ private struct SystemPictureInPictureHost: UIViewRepresentable {
...
         func pictureInPictureControllerDidStopPictureInPicture(_ pictureInPictureController: AVPictureInPictureController) {
             updateSourcePlaceholder(visible: false, animated: true)
             detachRenderer()
+            isStoppingForForegroundActivation = false
+        }
+
+        private func registerForegroundObservers() {
+            let notifications = [
+                UIApplication.willEnterForegroundNotification,
+                UIApplication.didBecomeActiveNotification
```
agatx: **Registering both `willEnterForeground` and `didBecomeActive` double-fires the stop on every foreground return.** On a normal background→foreground transition both notifications post in sequence, so `stopPictureInPictureForForegroundActivation()` runs twice. The second pass is mostly idempotent via the `isPictureInPictureActive` guard, but it re-issues `stopPictureInPicture()` on an already-stopping controller, which can drop the `didStop` callback feeding the stuck-flag issue above. `didBecomeActive` is also broader than "returns to the foreground" (the CHANGELOG/docs wording) since it posts for in-place reactivations. `willEnterForegroundNotification` alone covers the real background-return case — consider dropping the `didBecomeActive` observer.

samples/ios/Resources/Info.plist
```diff
@@ -28,6 +28,11 @@
 	<string>Serenada Sample needs microphone access for audio calls.</string>
 	<key>UIApplicationSupportsIndirectInputEvents</key>
 	<true/>
+	<key>UIBackgroundModes</key>
+	<array>
+		<string>audio</string>
+		<string>voip</string>
```
agatx: **`voip` background mode is declared with no PushKit/CallKit anywhere in the repo.** Camera PiP for video calls only needs `audio`; `voip` is for `PKPushRegistry`-driven VoIP wake-ups, and Apple has historically rejected apps that declare it without a registered push registry. This mirrors the production app's `Info.plist` (which also ships `[audio, voip]` with no PushKit), so it's pre-existing rather than introduced here — but the sample now teaches integrators a config that can trip App Store review. If `voip` isn't actually used, `audio` alone is the correct minimal set for system PiP.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
```diff
@@ -557,12 +561,39 @@ private struct SystemPictureInPictureHost: UIViewRepresentable {
...
+            ]
+            foregroundObservers = notifications.map { name in
+                NotificationCenter.default.addObserver(forName: name, object: nil, queue: .main) { [weak self] _ in
+                    Task { @MainActor [weak self] in
+                        self?.stopPictureInPictureForForegroundActivation()
+                    }
+                }
```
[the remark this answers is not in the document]
agatx: Addressed in 82d6d1db by removing the nested Task hop and keeping a single weak self capture in the observer callback.

docs/sdk/sdk-integration-ios.md
```diff
@@ -90,7 +90,7 @@ SerenadaCallFlow(
 )
 ```
 
-When enabled on a supported iOS device, the SDK keeps the active large video feed or avatar visible in system PiP and lets the system return control bring the user back to the app. The SDK enables multitasking camera access on its capture sessions when iOS reports support, so local camera video can continue in PiP on supported devices. iOS does not support custom in-window PiP actions for video calls, so End Call remains available only after returning to the app.
+When enabled on a supported iOS device, the SDK keeps the active large video feed or avatar visible in system PiP and lets the system return control bring the user back to the app. As the app becomes active again, the SDK stops the system PiP session and restores the inline call UI. The SDK enables multitasking camera access on its capture sessions when iOS reports support, so local camera video can continue in PiP on supported devices. iOS does not support custom in-window PiP actions for video calls, so End Call remains available only after returning to the app.
```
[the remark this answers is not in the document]
agatx: Addressed in 82d6d1db by rewriting the sentence to "lets the system return the user to the app".

docs/sdk/sdk-customization.md
```diff
@@ -73,7 +73,7 @@ Provider mode does not expose Serenada server helpers. These APIs require `serve
...
-| `systemPictureInPictureEnabled` | Bool | `false` | Android and iOS. Enables system Picture-in-Picture for waiting and active calls on supported mobile platforms. Android host apps must also set activity PiP manifest flags; iOS host apps must include background audio/VoIP modes. The SDK opts iOS capture sessions into multitasking camera access when supported and closes Android PiP by finishing the activity if the call ends while still in PiP. Custom PiP actions are not used, so call controls remain available after returning to the app. |
+| `systemPictureInPictureEnabled` | Bool | `false` | Android and iOS. Enables system Picture-in-Picture for waiting and active calls on supported mobile platforms. Android host apps must also set activity PiP manifest flags; iOS host apps must include background audio/VoIP modes. The SDK opts iOS capture sessions into multitasking camera access when supported, stops iOS system PiP as the app returns foreground, and closes Android PiP by finishing the activity if the call ends while still in PiP. Custom PiP actions are not used, so call controls remain available after returning to the app. |
```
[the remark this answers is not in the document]
agatx: Addressed in 82d6d1db: docs now say "returns to the foreground".

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
agatx: Addressed in 82d6d1db: update(...) now gates placeholderVisible on !isStoppingForForegroundActivation, so SwiftUI refreshes cannot re-show the placeholder during the foreground stop.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
agatx: Addressed in 82d6d1db: the flag is reset in cleanup() and in both PiP start callbacks, so it cannot leak into a later stop.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
agatx: Addressed in 82d6d1db: added a deinit observer removal fallback. Because deinit is nonisolated, it removes the token directly rather than calling the main-actor helper.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
agatx: Addressed in 82d6d1db: replaced Task { @MainActor } with MainActor.assumeIsolated on the already-main NotificationCenter queue and removed the redundant weak capture.

client-ios/SerenadaCallUI/Sources/SystemPictureInPicture.swift
agatx: Addressed in 82d6d1db: now observes only UIApplication.willEnterForegroundNotification, with an in-flight guard for idempotence.

samples/ios/Resources/Info.plist
agatx: Addressed in 82d6d1db: removed voip from the sample Info.plist/project.yml and docs now recommend only the background audio mode for iOS video PiP.

PR #134, opened 2026-06-25
[codex] Add independent screen share UI and fix iOS remote video

agatx:
## Summary

Adds the UI and host-app integration for independent screen-share/content streams on top of the core SDK/server support in PR #135, and fixes the iOS remote-video regression found during device validation.

## Scope

- Renders screen share as first-class content in web, Android Compose, and iOS SwiftUI call surfaces.
- Adds stream-keyed filmstrip/spotlight behavior for active screen shares.
- Updates host app plumbing to enable and pass through independent-content support.
- Improves Frontline remote screen-share viewing with framed/full-screen mode.
- Adds full-screen pinch-to-zoom and pan for remote screen shares on iOS and Android.
- Preserves fit-style rendering for remote screen shares, including in full-screen mode.
- Fixes iOS remote camera rendering when libwebrtc delivers legacy nil-transceiver stream callbacks after valid Unified Plan camera callbacks.
- Replays iOS remote renderer registrations when capability negotiation recreates a peer slot.

## Stack

Stacked on PR #135 (`independent-screen-share-core`). Review that PR first for protocol, SDK, media, and server support.

## Validation

Validation was run on the full stacked branch before splitting:

- `git diff --check`
- `./gradlew :serenada-call-ui:testDebugUnitTest --tests app.serenada.callui.ContentRenderingTest`
- `xcodegen generate`
- `xcodebuild -project SerenadaiOS.xcodeproj -scheme SerenadaiOS -destination platform=iOS Simulator,id=CA6FFD90-98C3-4580-BA33-11A912FC0BA9,arch=arm64 build`
- Installed and manually validated the latest Frontline screen-share UX on available iOS and Android devices.

Additional validation after the iOS remote-video fix:

- `git diff --check`
- `xcodebuild -project SerenadaiOS.xcodeproj -scheme SerenadaiOS -destination platform=iOS Simulator,id=CA6FFD90-98C3-4580-BA33-11A912FC0BA9 -only-testing:SerenadaiOSTests/RemoteVideoTransceiverRoutingTests -only-testing:SerenadaiOSTests/IndependentContentMediaTests/testCapabilityFlipReplaysRemoteRendererRegistrations test`
- `xcodebuild -project SerenadaiOS.xcodeproj -scheme SerenadaiOS -destination platform=iOS Simulator,id=CA6FFD90-98C3-4580-BA33-11A912FC0BA9 -only-testing:SerenadaiOSTests test`
- Patched iOS build deployed to attached iPad and iPhone; user confirmed iPad/iPhone remote video is fixed.

Co-authored by Codex (GPT-5 default)


agatx:
@codex please review

agatx:
@codex please review again

client-android/serenada-core/src/main/java/app/serenada/core/SerenadaSession.kt
```diff
@@ -1595,6 +1779,178 @@ class SerenadaSession internal constructor(
...
 
+    /**
+     * Store per-participant capabilities/media policy and replay any persisted
+     * content state carried in `joined`/`room_state` (e.g. for a peer that was
+     * already sharing when we reconnected). Reuses the same supersede-by-revision
+     * rule so a live `content_state` is never overwritten by an older snapshot.
+     */
+    private fun seedRemoteContentFromRoomState(roomState: RoomState) {
+        val myCid = clientId
+        for (participant in roomState.participants) {
```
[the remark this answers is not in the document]
agatx: Resolved as outdated after the PR split/rebase. The UI PR no longer owns the Android core reconnect snapshot path; content-state pruning for authoritative snapshots lives in the core PR (#135), where absent CIDs are removed from tracked content state.

client/src/pages/CallRoom.tsx
```diff
@@ -286,6 +286,10 @@ const CallRoom: React.FC = () => {
             serverHost: getConfiguredServerHost(),
             logger: new ConsoleSerenadaLogger(),
             turnsOnly,
+            // TEST ENABLE: independent screen-share content stream, for manual
+            // validation. SDK library default is false; revert or gate behind the
+            // legacy-receiver interop gate before production rollout.
+            enableIndependentContentVideo: true,
```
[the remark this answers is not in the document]
agatx: Intentional for this bundled host app. The feature is meant to ship enabled here while the headless SDK default remains false for external integrators. In eb26f81e I replaced the test-enable/rollout-disable comment with an explicit bundled-app opt-in constant/comment rather than turning the feature off.

client-android/serenada-call-ui/src/main/java/app/serenada/callui/CallScreen.kt
```diff
@@ -191,6 +204,56 @@ internal fun CallScreen(
...
+    // INDEPENDENT content stream (local or remote). In that case the whole stage
+    // switches to a single filmstrip+spotlight where EVERY stream is its own tile
+    // keyed {cid, kind} — a camera tile per camera-on participant and a content
+    // tile per sharer (incl the local user's own screen), so a sharer's camera and
+    // screen are two EQUAL peer tiles (not a camera-over-content PIP). This engages
+    // for a 1:1 + share exactly like multi-party. Legacy / flag-off content (no
+    // independent stream) is NOT stream-keyed and keeps the existing
+    // participant-keyed content path (byte-identical). Audio-only receivers resolve
+    // no content here, so the flag is false for them.
+    val streamKeyedStageActive = contentScene.all.any { it.mode == ContentMode.INDEPENDENT }
```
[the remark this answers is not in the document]
agatx: Fixed in eb26f81e. Android CallScreen now routes snapshot selection through pinnedStageTile while the stream-keyed stage is active, returns a source only for pinned camera tiles, and ignores content tiles. Added JVM coverage for camera/content/video-off cases in ContentRenderingTest.

client/packages/react-ui/src/SerenadaCallFlow.tsx
```diff
@@ -867,19 +901,227 @@ export const SerenadaCallFlow: React.FC<CallFlowProps> = ({
...
+                ? { cid: remoteContentState.cid, contentType: remoteContentState.contentType }
+                : null,
+            accessors: {
+                getLocalContentStream: () => session?.getLocalContentStream() ?? null,
+                getRemoteContentStream: (cid) => session?.getRemoteContentStream(cid),
+            },
+            // Audio-only receivers (videoMediaEnabled=false) never negotiated
+            // content receive and must suppress all content UI.
+            localVideoMediaEnabled: videoEnabledConfig,
+            remoteContentOrder,
```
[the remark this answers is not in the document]
agatx: Fixed in eb26f81e. React UI now has a videoMediaEnabled config that mirrors SerenadaConfig.videoMediaEnabled and is distinct from the UI camera-control toggle. The content resolver uses that policy for strict audio-only suppression, and URL-first sessions pass it into SerenadaCore.

client/packages/react-ui/src/SerenadaCallFlow.tsx
```diff
@@ -1168,19 +1441,156 @@ export const SerenadaCallFlow: React.FC<CallFlowProps> = ({
...
+                                            return (
+                                                <div
+                                                    key={tile.id}
+                                                    className="video-stage-tile"
+                                                    style={tileStyle}
+                                                    data-testid="call.contentLoading"
+                                                    role="button"
+                                                    tabIndex={0}
+                                                    onPointerUp={(event) => { event.stopPropagation(); togglePin(); }}
+                                                >
```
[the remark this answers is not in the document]
agatx: Fixed in eb26f81e. The content-loading tile now handles Enter and Space, prevents default scrolling for Space, stops propagation, and toggles the same pin action as pointer input.

client/src/pages/CallRoom.tsx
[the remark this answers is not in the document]
agatx: Intentional for this bundled web host. The new feature is meant to be enabled in this app; the SDK library default remains false. eb26f81e updates the stale test/rollout comment to an explicit bundled-app opt-in constant/comment instead of gating it off.

client-android/app/src/main/java/app/serenada/android/call/CallManager.kt
```diff
@@ -150,6 +150,10 @@ class CallManager(context: Context) : RoomWatcherDelegate {
                 isHdVideoExperimentalEnabled = settingsStore.isHdVideoExperimentalEnabled,
                 transports = transports,
                 proximityMonitoringEnabled = true,
+                // TEST ENABLE: independent screen-share content stream, for manual
+                // validation. SDK library default is false; revert or gate behind the
+                // legacy-receiver interop gate before production rollout.
+                enableIndependentContentVideo = true,
```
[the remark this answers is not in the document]
agatx: Intentional for this bundled Android host. eb26f81e keeps the feature enabled but replaces the stale test-enable wording with an explicit bundled-app opt-in constant/comment, so the SDK default can remain false for integrators.

client-ios/Sources/Core/Call/CallManager.swift
```diff
@@ -425,6 +426,21 @@ final class CallManager: ObservableObject {
                 serverHost: host,
                 defaultAudioEnabled: settingsStore.isDefaultMicrophoneEnabled,
                 defaultVideoEnabled: frontline ? false : settingsStore.isDefaultCameraEnabled,
+                // TEST ENABLE: independent screen-share content stream, for manual
+                // validation. SDK library default is false; revert or gate behind the
+                // legacy-receiver interop gate before production rollout. (Standard
+                // CallScreen renders the independent content tile; FrontlineCallScreen
+                // keeps its single-content presentation.)
+                enableIndependentContentVideo: true,
```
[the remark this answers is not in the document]
agatx: Intentional for this bundled iOS host. eb26f81e keeps the feature enabled but replaces the stale test-enable wording with an explicit bundled-app opt-in comment; the SDK default remains false for external integrators.

client-ios/Sources/App/RootView.swift
```diff
@@ -119,7 +119,13 @@ struct RootView: View {
...
                             callManager.dismissActiveCall()
                         },
-                        onDismiss: { callManager.dismissActiveCall() }
+                        onDismiss: { callManager.dismissActiveCall() },
+                        // The session is built with enableIndependentContentVideo=true
+                        // (CallManager), so the prebuilt UI must echo the flag or it
+                        // resolves every share as LEGACY and never renders the remote
+                        // screen-share content tile (the SDK still negotiates + receives
+                        // the content track; only the UI presentation was gated off).
+                        independentContentVideo: true
```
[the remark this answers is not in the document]
agatx: Fixed in eb26f81e. The iOS host UI no longer hard-codes a separate literal; it now reads CallManager.independentContentVideoEnabled, keeping the UI flag in lockstep with the session opt-in.

client-android/app/src/main/java/app/serenada/android/ui/SerenadaAppRoot.kt
```diff
@@ -349,6 +349,10 @@ fun SerenadaAppRoot(
                         SerenadaCallFlow(
                             session = session,
                             initialRemoteVideoFitCover = callManager.isRemoteVideoFitCover,
+                            // Session is built with enableIndependentContentVideo=true
+                            // (CallManager); echo it so the UI renders the remote
+                            // screen-share content tile instead of the legacy path.
+                            independentContentEnabled = true,
```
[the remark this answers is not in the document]
agatx: Fixed in eb26f81e. The Android host UI now imports and uses the same INDEPENDENT_CONTENT_VIDEO_ENABLED constant as CallManager, so UI and session opt-in stay in sync while remaining intentionally enabled for the bundled app.

client-android/app/src/main/java/app/serenada/android/service/CallService.kt
```diff
@@ -34,7 +34,11 @@ class CallService : Service() {
                 mediaProjectionForegroundActive = includeMediaProjection
             } else {
                 startForeground(NOTIFICATION_ID, buildNotification(roomId, roomName))
-                mediaProjectionForegroundActive = false
+                // Pre-Q has no foreground-service-TYPE gating for MediaProjection, so a
+                // running foreground service is sufficient. Mark ready when this start
+                // was for a screen share, else startScreenShareWhenForegroundReady never
+                // passes its gate and the share silently does nothing on old Android.
+                mediaProjectionForegroundActive = includeMediaProjection
```
agatx: **🔴 Android screen share can silently fail to start (and tear down an active capture)** (plausible)

This PR now sets `mediaProjectionForegroundActive = includeMediaProjection` on **both** branches, so a `CallService.start()` call with the default `includeMediaProjection=false` resets the flag to false.

Flow: user taps share → `startScreenShare()` calls `start(includeMediaProjection=true)` and polls `startScreenShareWhenForegroundReady` for that flag. The SDK then emits a new `session.state` (localContent becomes non-null) → `handleSdkSessionState` → the `else ->` branch at `CallManager.kt:236` → `CallService.start()` with the default `false` → flag flips back to false. The poll never passes its gate (~15×50ms), gives up, and `session.startScreenShare(intent)` is never invoked — the share silently does nothing.

On Q+ the same restart re-calls `startForeground` **without** `FOREGROUND_SERVICE_TYPE_MEDIA_PROJECTION` while capture is active, which can tear down an in-progress share.

Consider making routine `start()` calls not clear the projection flag (only an explicit stop should), or have the `else ->` re-emit path preserve `includeMediaProjection` when a share is starting/active.

client-ios/SerenadaCallUI/Sources/CallScreen.swift
```diff
@@ -894,7 +991,7 @@ struct CallScreenView: View {
                 }
             }
 
-            if config.screenSharingEnabled {
+            if config.screenSharingEnabled && screenShareAvailable {
```
agatx, line 1004: **🟠 iOS screen-share button silently disappears for integrators** (confirmed)

Button visibility was narrowed from `if config.screenSharingEnabled` to `if config.screenSharingEnabled && screenShareAvailable` (same on `FrontlineCallScreen.swift:1010`). `screenShareAvailable` is false whenever `session.screenShareMode == .disabled`.

A host that set `config.screenSharingEnabled = true` but didn't configure a broadcast `screenShareMode` previously showed a button that no-op'd; after this upgrade the button vanishes entirely with no config change on their side. If hiding is intended, that's a documented behavior change worth calling out in the changelog; otherwise consider keeping the button visible (disabled/with guidance) when `screenSharingEnabled` is set.

client-ios/SerenadaCallUI/Sources/CallScreen.swift
```diff
@@ -697,6 +779,21 @@ struct CallScreenView: View {
     }
 
     private var currentSnapshotSource: SnapshotSource? {
+        // Stream-keyed content stage: the spotlight is a specific stream. A camera
+        // tile in the spotlight is snapshot-able; a content (screen-share) tile is
+        // not, so hide the shutter there. `stageSpotlightId` already folds in the
+        // pin, so this targets exactly what is large.
+        if streamKeyedStageActive {
```
agatx: **🟠 Snapshot shutter disappears when a screen share is spotlighted (iOS)** (confirmed)

The snapshot source in the stream-keyed stage returns `nil` when the spotlight tile is a content (screen-share) tile. The legacy multi-party `snapshotSource` always resolved the pinned/large preview to a Local/Remote camera source.

Net effect: while someone is presenting (the common case), the shutter is hidden, so a user who taps to capture the call sees no shutter and can't grab a snapshot of the presentation filling their screen. Consider resolving the snapshot source to the spotlighted tile regardless of whether it's a camera or content tile.

client/packages/react-ui/src/utils/contentRendering.ts
```diff
@@ -0,0 +1,588 @@
...
+    if (!local) return null;
+    // Audio-only receivers never present content UI. Sharing is also blocked by
+    // the SDK for the local user in this mode, so there is nothing to show.
+    if (!localVideoMediaEnabled) return null;
+    if (!localContentActive(local)) return null;
+
+    // Only a real screen share is independent content; a camera-framing local
+    // content_state (world/composite) has no content track and renders via the
+    // camera/legacy path (mirrors the remote gate + mobile's isScreenShareContent).
+    const independentStream = accessors.getLocalContentStream() ?? null;
```
agatx: **🟠 Content independent-vs-legacy detection diverges across platforms (root of the web camera-leak)** (confirmed)

The shared content-resolution decision keys on different root signals per platform: web keys "independent mode" on `getLocalContentStream()`/`getRemoteContentStream()` returning non-null, while iOS/Android key on `independentContentEnabled + supportsIndependentContentVideo`. (Also at `contentRendering.ts:230`, `:296`, and `ContentRendering.swift:363`.)

Because the detection mechanism differs at the root, the three platforms drift rather than behave identically — every future change to content gating must be hand-mirrored across three incompatible paths, and this divergence already produces the brief camera-as-screen-share leak on web. Worth converging on one keying contract (all platforms key on the same capability/stream signal).

client/packages/react-ui/src/utils/contentRendering.ts
```diff
@@ -0,0 +1,588 @@
...
+    }
+    return false;
+}
+
+/** True when a stream carries at least one live (non-ended), enabled video track. */
+function streamHasLiveVideo(stream: MediaStream | null | undefined): boolean {
+    if (!stream) return false;
+    const tracks = stream.getVideoTracks();
+    if (tracks.length === 0) return false;
+    return tracks.some((t) => t.readyState !== 'ended');
```
agatx: **🟠 `streamHasLiveVideo` ignores `track.enabled` despite its doc** (confirmed)

The doc comment says "at least one live (non-ended), **enabled** video track", but the implementation only checks `t.readyState !== 'ended'` and never checks `t.enabled` (also at `:434`). A present-but-disabled (muted) content video track counts as live.

So if an independent screen-share content track is delivered with its video track disabled/muted, `hasLiveTrack` is true → the content tile renders the blank/frozen muted stream with `loading:false` instead of the connecting spinner the receiver-side-hold design intends → viewer sees a black/stale content spotlight instead of a "connecting" tile. Add the `&& t.enabled` check to match the doc and the intended hold behavior.

client/packages/react-ui/src/SerenadaCallFlow.tsx
```diff
@@ -867,19 +902,227 @@ export const SerenadaCallFlow: React.FC<CallFlowProps> = ({
...
+            const observer = new ResizeObserver(updateViewportSize);
+            observer.observe(node);
+            return () => observer.disconnect();
         }
-        if (remoteContentState) {
-            return { type: remoteContentState.contentType, ownerParticipantId: remoteContentState.cid, aspectRatio: null };
+
+        window.addEventListener('resize', updateViewportSize);
+        return () => window.removeEventListener('resize', updateViewportSize);
+    }, [useContentStageLayout]);
```
agatx: **🟡 Stage ResizeObserver can hold a stale viewport size** (plausible)

The stage-viewport ResizeObserver effect depends only on `[useContentStageLayout]`. If the content stage stays mounted (boolean stays true) but `stageViewportRef`'s DOM node is replaced across a re-render (e.g. a conditional subtree swap between the legacy multi-party and independent stage), the effect doesn't re-attach to the new node and `stageViewportSize` keeps its old value.

`computeStageLayout` then lays out tiles against wrong dimensions → clipped, zero-size, or misplaced video tiles. Consider a ref callback or including the node identity so the observer re-attaches on remount.

client/packages/react-ui/src/SerenadaCallFlow.tsx
```diff
@@ -594,30 +629,6 @@ export const SerenadaCallFlow: React.FC<CallFlowProps> = ({
...
-            return () => observer.disconnect();
-        }
-
-        window.addEventListener('resize', updateViewportSize);
-        return () => window.removeEventListener('resize', updateViewportSize);
-    }, [isMultiParty]);
-
     const clearIdleHide = useCallback(() => {
         if (idleTimeoutRef.current !== null) {
             window.clearTimeout(idleTimeoutRef.current);
```
agatx, line 637: **🟡 Web shows the wrong "waiting" message for a pre-join screen share** (confirmed)

The independent-share waiting badge reuses the generic `waitingForOther` ("Waiting for the other person to join") instead of a dedicated content-waiting string (also at `:572`). Android and iOS added dedicated strings (`CallContentWaitingForParticipants` / `callContentWaitingForParticipants`) for exactly this state.

A web user who starts a screen share before anyone joins sees "Waiting for the other person to join" on the share preview — the wrong message, and inconsistent with the "Sharing, waiting for participants" shown on mobile. Add a matching web string for parity.

> _Anchored here; the code this refers to is at `SerenadaCallFlow.tsx:639`._

client-ios/SerenadaCallUI/Sources/CallScreen.swift
```diff
@@ -1012,6 +1109,64 @@ struct CallScreenView: View {
...
+    /// Never true flag-off / legacy (the resolver yields no `.independent`
+    /// content), so the legacy/no-content paths stay byte-identical.
+    private var hasIndependentContent: Bool {
+        stageContentAll.contains { $0.mode == .independent }
+    }
+
+    /// The full stream-keyed tile list for the stage (one tile per camera-on
+    /// participant + one per independent sharer, keyed {cid, kind}). Empty unless
+    /// an independent content stream is present, so the legacy paths are untouched.
+    private var stageTiles: [StageTile] {
```
agatx: **🔵 Uncached per-body recomputation of stage tiles (iOS)** (confirmed)

`stageTiles` (`deriveStageTiles` over cameras + `contentScene.all`), `stageSpotlightId` (`pickStageSpotlightTileId`), and `streamKeyedStageActive` are plain computed vars referenced ~22× across the body and `onChange` modifiers. Swift caches nothing, so each reference re-runs `deriveStageTiles`/the resolver scan — the stage rebuilds its tile list and re-scans participants many times per redraw, while content is being shared.

Web and Android memoize each of these once (`useMemo`/`remember`); also applies to `FrontlineCallScreen.swift:271`, `CallScreen.swift:1570`, and the Android `ContentRendering.kt:150` call site. Compute each once and store to match the memoized paths.

client-ios/SerenadaCallUI/Sources/ContentRendering.swift
```diff
@@ -0,0 +1,732 @@
...
+///     experience byte-identical.
+struct ResolvedContentSource: Equatable {
+    let type: ContentType
+    let ownerCid: String
+    let mode: ContentMode
+}
+
+// MARK: - Local / remote content activity + type
+
+private func localContentType(_ local: ContentLocalParticipant) -> ContentType {
```
agatx, line 182: **🔵 Content-resolution module maintained in triplicate** (plausible)

The full resolver (`localContentActive/Type`, `remoteContentActive/Type`, `resolveLocalContent`, `resolveRemoteContents`, `pickPrimaryContent`, `deriveStageTiles`, `pickStageSpotlightTileId`, tile-id encode/parse) is reimplemented across `contentRendering.ts`, `ContentRendering.kt`, and `ContentRendering.swift` with matching shapes and parallel test suites.

This is the established headless-per-platform pattern, but the gating logic here (screen-share-only independent tiles, receiver-side hold, most-recent-primary ordering) is non-trivial enough that a fix to one platform not mirrored will produce cross-platform content-tile divergence. Consider a shared conformance fixture (like `layout_conformance_v1.json`) so the three copies are checked for parity rather than hand-kept.

client/packages/react-ui/src/utils/contentRendering.ts
```diff
@@ -0,0 +1,588 @@
...
+    const independentRemotes = remotes.filter((r) => r.mode === 'independent');
+    if (independentRemotes.length > 0) {
+        if (remoteContentOrder && remoteContentOrder.length > 0) {
+            // Most-recently-active is LAST in order. Walk back to front.
+            for (let i = remoteContentOrder.length - 1; i >= 0; i -= 1) {
+                const match = independentRemotes.find((r) => r.ownerId === remoteContentOrder[i]);
+                if (match) return match;
+            }
+        }
+        return independentRemotes[independentRemotes.length - 1];
```
agatx, line 306: **🔵 `deriveStageTiles` accepts a `cameraOn` field it never reads** (plausible)

Each `StageCameraParticipant` carries a `cameraOn` field, but `deriveStageTiles` (web/iOS/Android) never reads it — it unconditionally emits a camera tile per participant, and callers compute `cameraOn` (e.g. `remoteParticipantMap.get(cid)?.cameraEnabled !== false`) for no effect.

No direct runtime failure, but the dead field masks intent: a future change relying on `cameraOn` to omit camera-off tiles would silently do nothing, and the camera-off filmstrip behavior depends entirely on the downstream `if (!stream) return null` render guard. Either wire `cameraOn` into the tile decision or drop it.

client-android/app/src/main/java/app/serenada/android/service/CallService.kt
agatx: Validated and fixed in 6f09f5d. Normal CallService refreshes now preserve an already-requested media-projection foreground type, while stop/failure paths explicitly clear it so the service can downgrade after sharing ends. Verified with :app:compileDebugKotlin.

client-ios/SerenadaCallUI/Sources/CallScreen.swift
agatx: Validated. The iOS button visibility is intentionally gated by both integrator config and SDK availability; the feature stays enabled in bundled apps. I updated the docs to call out the iOS isScreenShareAvailable / screenShareMode availability gate and removed stale native README text that said independent content was not yet enabled. Fixed in 6f09f5d.

client-ios/SerenadaCallUI/Sources/CallScreen.swift
agatx: Validated and fixed in 6f09f5d. The standard iOS shutter no longer disappears only because a content tile is spotlighted; it now targets the spotlight owner camera source when that camera is enabled. The current snapshot API remains camera-source based, so it does not claim to capture the content stream itself. Verified with the SerenadaCallUI and SerenadaiOS simulator builds.

client/packages/react-ui/src/utils/contentRendering.ts
agatx: Validated and fixed in 6f09f5d. Web content resolution now uses the same contract as mobile: local independentContentEnabled plus per-remote supportsIndependentContentVideo, instead of stream presence alone. I also exposed the existing core capability accessors on the web session handle and added a regression test for a non-capable peer with an apparent content stream staying LEGACY. Verified with the React UI contentRendering test and client build.

client/packages/react-ui/src/utils/contentRendering.ts
agatx: Validated and fixed in 6f09f5d. streamHasLiveVideo now treats disabled video tracks as not live media, and the web resolver test suite covers the disabled-track loading case. Verified with the React UI contentRendering test and client build.

client/packages/react-ui/src/SerenadaCallFlow.tsx
agatx: Validated and hardened in 6f09f5d. The stage ResizeObserver is now attached via a callback-ref backed DOM node state, so the effect depends on the actual mounted viewport node and reattaches after remounts. Verified with the client build.

client/packages/react-ui/src/SerenadaCallFlow.tsx
agatx: Validated and fixed in 6f09f5d. Web now has a dedicated contentWaitingForParticipants string and uses it only for local screen-share waiting badges, matching the mobile copy while leaving the normal waiting-room text unchanged. Verified with the client build.

client-ios/SerenadaCallUI/Sources/CallScreen.swift
agatx: Validated and addressed in 6f09f5d. Standard iOS now snapshots stream-keyed stage derivation once per body render for body branch decisions, spotlight PiP selection, and stale-pin cleanup. Android Frontline also memoizes the stage tile and spotlight derivations with remember. Verified with Android call-ui unit tests, Android app compile, SerenadaCallUI build, and SerenadaiOS build.

client-ios/SerenadaCallUI/Sources/ContentRendering.swift
agatx: Validated as a real maintainability risk. I did not introduce a new cross-language fixture in this PR because the current repo pattern is parallel platform unit suites, and adding a shared fixture runner across TS/Kotlin/Swift would be broader infrastructure work. For this review I tightened the web parity contract to match mobile and added regression coverage for the divergence that caused the leak. Fixed/explained in 6f09f5d.

client/packages/react-ui/src/utils/contentRendering.ts
agatx: Validated and fixed in 6f09f5d. StageCameraParticipant no longer carries the unused cameraOn field across web, Android, or iOS. The behavior is unchanged: camera-off participants still get avatar camera tiles in the stream-keyed stage by design. Verified with web resolver tests, Android call-ui tests, Android app compile, and iOS builds.

PR #135, opened 2026-06-25
[codex] Add independent screen share core support

agatx:
## Summary

Adds the SDK, protocol, media, and server foundation for independent screen-share/content streams across Serenada.

## Scope

- Adds content/media state and signaling payload support for independent screen-share streams.
- Adds web core, Android core, and iOS core sender/receiver routing for content tracks.
- Adds iOS broadcast IPC/runtime screen-share mode support and delayed signaling until capture starts.
- Adds server signaling coverage and protocol/docs updates.
- Includes regression coverage for content state, content media, transceiver routing, diagnostics, and related screen-share lifecycle cases.

## Stack

This is the base PR for the screen-share split. The UI/host rendering changes are stacked in PR #134 on top of this branch.

## Validation

Validation was run on the full stacked branch before splitting:

- `git diff --check`
- `./gradlew :serenada-call-ui:testDebugUnitTest --tests app.serenada.callui.ContentRenderingTest`
- `xcodegen generate`
- `xcodebuild -project SerenadaiOS.xcodeproj -scheme SerenadaiOS -destination 'platform=iOS Simulator,id=CA6FFD90-98C3-4580-BA33-11A912FC0BA9,arch=arm64' build`
- Installed and manually validated the latest Frontline screen-share UX on available iOS and Android devices.

Co-authored by Codex (GPT-5 default)


client-ios/BroadcastUpload/ScreenShareSampleHandler.swift
```diff
@@ -0,0 +1,10 @@
+import SerenadaBroadcastExtensionSupport
```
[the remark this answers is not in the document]
agatx: Addressed in https://github.com/agatx/serenada/pull/134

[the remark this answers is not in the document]
agatx: Fixed in 4b28b7de. The broadcast extension wiring now lives in the core PR: project.yml adds SerenadaBroadcastExtensionSupport to the app and extension targets, points the extension principal at ScreenShareSampleHandler, includes the app group identifier/API-only setting, and the checked-in Xcode project was regenerated. Verified with `xcodebuild -project client-ios/SerenadaiOS.xcodeproj -scheme SerenadaBroadcast -destination 'generic/platform=iOS' CODE_SIGNING_ALLOWED=NO build`.

client/packages/core/src/media/MediaEngine.ts
```diff
@@ -137,6 +218,11 @@ export class MediaEngine {
     private deferInitialAnswer: boolean;
     private logger?: SerenadaLogger;
     private offerSequence = 0;
+    // Per-session monotonic generation marker for outgoing `content_state`.
+    // Incremented on every send (start/stop) so receivers can order
+    // presentation-state changes within this (cid, sid). Starts at 0; the first
+    // send carries `revision: 1`.
+    private contentRevision = 0;
```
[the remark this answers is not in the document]
agatx: Fixed in 4b28b7de. MediaEngine now exposes content revision seeding, and SerenadaSession seeds it from the recovered local participant content state in joined/room_state snapshots so the next local content_state send advances past the persisted revision. Added a web core regression test in SerenadaSessionContentState.test.ts.

docs/serenada_protocol_v1.md
```diff
@@ -137,6 +141,30 @@ Join a room.
...
+**Capabilities / mediaPolicy allowlist**
+- The server stores only **allowlisted** keys:
+  - `capabilities`: `trickleIce`, `maxParticipants`, `independentContentVideo`.
+  - `mediaPolicy`: `videoMediaEnabled`.
+- Any other key under `capabilities` or `mediaPolicy` is **dropped** (not stored,
+  not forwarded). Opaque pass-through of arbitrary future keys is intentionally
+  not supported in this protocol version.
+- Allowlisted values are stored per participant and re-applied on every join,
+  including reconnect; a reconnect that omits the objects keeps the previously
+  stored values.
```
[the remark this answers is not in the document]
agatx: Fixed in 4b28b7de. The protocol spec now matches the server behavior: allowlisted capabilities/mediaPolicy are taken from the latest successful join, and omitted objects on a later join/reconnect clear the stored value so clients apply defaults.

server/signaling.go
```diff
@@ -1600,6 +1701,15 @@ func applyContentStateUpdate(room *Room, senderCID string, payload map[string]in
...
 	}
+	// revision is the sender's per-(cid,sid) monotonic counter. The server keeps
+	// it verbatim (latest-wins, like the rest of content_state) so a reconnecting
+	// peer learns the sender's current revision via room_state. The server does
+	// NOT compare or order revisions itself — that comparison is a receiver-side
+	// rule (a stale active:false that should not override a newer active:true is
+	// the receiver's responsibility to discard). JSON numbers decode as float64.
+	if revision, ok := payload["revision"].(float64); ok {
+		state.Revision = int64(revision)
+	}
```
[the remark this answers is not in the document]
agatx: Fixed in 4b28b7de. The server now persists revision only when the JSON number is a non-negative safe integer; fractional, negative, and unsafe values are ignored. Added Go coverage in TestContentStateRevisionPersistsOnlySafeIntegers.

client/packages/core/src/types.ts
```diff
@@ -197,6 +256,14 @@ export interface SerenadaConfig {
      * video are all disabled. Defaults to `true`.
      */
     videoMediaEnabled?: boolean;
+    /**
+     * Static capability gate for the independent screen-share content stream.
+     * When `true`, this build can negotiate a dedicated content video stream
+     * separate from the camera. Advertised to peers at join via
+     * `capabilities.independentContentVideo`. Immutable per session. Defaults to
+     * `false`. Phase 1: advertisement only; not yet consumed by media.
+     */
```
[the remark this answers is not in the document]
agatx: Fixed in 4b28b7de. The SerenadaConfig docstring now says enableIndependentContentVideo both advertises the capability and enables the independent content media path for capable peers, with legacy peers staying on the single-video fallback.

client-ios/SerenadaCore/Sources/SignalingProvider.swift
```diff
@@ -26,17 +26,27 @@ public struct JoinOptions: Equatable, Sendable {
     /// server-issued) — lets host applications correlate a participant to their
     /// own user identity (avatar lookup, telemetry).
     public var appPeerId: String?
+    /// Static build capability advertised in `join.capabilities`. Mirrors
+    /// `SerenadaConfig.enableIndependentContentVideo`. Phase 1: always `false`.
+    public var independentContentVideo: Bool
```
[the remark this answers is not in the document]
agatx: Fixed in 4b28b7de. The iOS JoinOptions comment now documents default false plus host opt-in, instead of saying the capability is always false.

README.md
```diff
@@ -226,6 +226,23 @@ In custom-provider mode:
 
 On web, provider-delivered peer messages are exposed through `session.onPeerMessage(...)`. Raw Serenada transport envelopes are no longer part of the public SDK surface.
 
+### Independent Content Video (Screen Share)
+
+Screen share is moving from the single camera-video path to its own independent **content** stream, so a capable peer can send camera and screen share at the same time. The SDK surface for this is in place and stable, but the media path is gated behind a config flag that is **off by default and not yet enabled** on any platform — flipping it is held until each platform's media + UI are ready and the legacy-receiver interop gate passes. With the flag off, behavior is exactly today's: screen share reuses the camera video path and replaces the camera track while sharing.
+
+- **Config flag** `enableIndependentContentVideo` (`SerenadaConfig`, default `false`). It gates one thing only: advertising `capabilities.independentContentVideo=true` at join. It does not change media behavior until the per-platform media engine, UI, and interop gate are all ready.
```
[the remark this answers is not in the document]
agatx: Fixed in 4b28b7de. README now describes the flag as off by default in the headless SDK but enabling independent content media for capable peers when a host app opts in.

client-ios/SerenadaCore/Sources/SerenadaConfig.swift
```diff
@@ -71,6 +105,8 @@ public struct SerenadaConfig: Equatable, @unchecked Sendable {
         self.defaultAudioEnabled = defaultAudioEnabled
```
agatx, line 105: **🔴 Critical regression — iOS screen sharing fully disabled** (CONFIRMED)

Screen-share gating moved from compile-time `#if BROADCAST_EXTENSION` to runtime `config.screenShareMode`, which defaults to `.disabled`. The old code set `screenShareExtensionBundleId` whenever `BROADCAST_EXTENSION` was defined (it was, on the host target). Now it returns nil unless the host passes `screenShareMode: .broadcast(...)`.

`CallManager.swift` (`SerenadaConfig(...)`, ~line 424) does **not** pass `screenShareMode`, so it defaults to `.disabled`, `isScreenShareAvailable` is false, and the new `guard isScreenShareAvailable else { return }` in `startScreenShare()` makes the share button a silent no-op. `project.yml` also removed `BROADCAST_EXTENSION` from `SWIFT_ACTIVE_COMPILATION_CONDITIONS`, so there is no fallback.

Net effect: after this PR, iOS users (and any unmigrated host) can no longer start a screen share. Migrate the host config to pass `screenShareMode: .broadcast(...)`.

client-ios/SerenadaCore/Sources/Call/ScreenShareCapturers.swift
```diff
@@ -139,11 +233,15 @@ final class BroadcastFrameReader: RTCVideoCapturer {
...
+        // the read (matching the writer's odd→even publish).
         let seqNo = ptr.load(fromByteOffset: BroadcastHeaderOffset.seqNo, as: UInt32.self)
+        guard seqNo & 1 == 0 else { return }
         guard seqNo != lastSeqNo else { return }
+        OSMemoryBarrier()
 
+        let generation = ptr.load(fromByteOffset: BroadcastHeaderOffset.generation, as: UInt32.self)
         let width = Int(ptr.load(fromByteOffset: BroadcastHeaderOffset.width, as: UInt32.self))
         let height = Int(ptr.load(fromByteOffset: BroadcastHeaderOffset.height, as: UInt32.self))
         let pixelFormat = ptr.load(fromByteOffset: BroadcastHeaderOffset.pixelFormat, as: UInt32.self)
```
agatx, line 247: **🔴 Crash risk — unbounded shared-memory over-read in `pollFrame`** (CONFIRMED)

`pollFrame` reads plane `width`/`height`/`bytesPerRow`/`pixelFormat` from the shared-memory header and uses them directly as `memcpy` source sizes and loop bounds (`plane0Size = plane0BytesPerRow * plane0Height`, copied from `dataStart`) with **no validation** that these fit within the mmap'd file size. The seqlock re-check happens only *after* the copy, so an even-seq but internally-inconsistent header is copied first and validated later.

If the reader samples a header mid-update where `seqNo` is even but plane dimensions are stale/oversized relative to the actual buffer (or a malformed/garbage region), the `memcpy` over-reads past the mapped region → out-of-bounds read → SIGSEGV/EXC_BAD_ACCESS in the in-call app while receiving a screen share.

Validate plane sizes against the mapped length before copying, and/or re-check the seqlock before trusting the dimensions.

> _Anchored here because `ScreenShareCapturers.swift:278` is outside the PR diff; the issue is at line 278._

client-ios/SerenadaCore/Sources/SerenadaSession.swift
```diff
@@ -1779,6 +2097,64 @@ public final class SerenadaSession: ObservableObject {
         mediaLivenessEmitCount += 1
```
agatx: **🟠 Cross-platform parity broken — `media_liveness` empty heartbeat** (CONFIRMED)

The web SDK was changed (`client/packages/core/src/SerenadaSession.ts:1639`, with an explicit comment) to emit an empty `media_liveness` heartbeat when `flowing` is empty, so the server can fast-evict suspended ghosts in the ~30s sweep. iOS (here) and Android (`SerenadaSession.kt:2107`) were **not** updated — `finalizeMediaLivenessEmit` still early-returns on empty `flowing`.

Result on iOS/Android: when all remote peers drop simultaneously, the lone survivor has no inbound flow, so it stops reporting entirely. The ghost peers and the survivor's "Reconnecting" UI linger until the 10-minute hard-eviction timer instead of clearing in ~30s. Port the empty-heartbeat emit to both mobile SDKs.

> _Anchored here because `SerenadaSession.swift:2094` is outside the PR diff; the issue is at line 2094._

server/signaling.go
```diff
@@ -1165,6 +1252,21 @@ func (h *Hub) handleJoin(c *Client, msg Message) {
 		p.PeerID = trimmed
 	}
 
+	// Store the allowlisted capabilities/mediaPolicy on every join (including
+	// reconnect) so the values stay current and are forwarded in
+	// joined/room_state. Each join is authoritative: assign the freshly parsed
+	// records verbatim, even when nil. Per the wire contract, an omitted
+	// capabilities/mediaPolicy (or an omitted allowlisted key) means "apply
```
agatx: **🟠 Capabilities/MediaPolicy cleared on reconnect** (CONFIRMED)

On every join (including token-recovery reconnect) the server now unconditionally overwrites `p.Capabilities`/`p.MediaPolicy` with the freshly parsed records, clearing them to nil when a (re)join omits those objects.

A third-party SDK that advertises `independentContentVideo:true` at initial join but omits the `capabilities` object on its reconnect join would have its stored capability cleared and forwarded as absent. Peers then re-derive that participant as legacy mid-call, stop negotiating its content track independently, and the screen share disappears until an explicit re-share. In-family SDKs always re-send capabilities so they're unaffected; the exposure is third-party clients against the documented "omitted = apply defaults" contract. Consider only overwriting when the field is present in the payload.

client-ios/SerenadaCore/Sources/SerenadaSession.swift
```diff
@@ -1032,12 +1174,110 @@ public final class SerenadaSession: ObservableObject {
...
+    ///
+    /// Revisionless edge case differs by caller and MUST stay identical to the
+    /// historical two-method behavior:
+    ///   - live (`isSnapshot: false`): a revisionless update is ALWAYS applied
+    ///     (legacy senders advance presentation state without revision gating);
+    ///   - snapshot (`isSnapshot: true`): a revisionless snapshot KEEPS the
+    ///     cache when one exists (a live message is the more authoritative
+    ///     recent source); with no cache it is adopted.
+    @discardableResult
+    private func applyRemoteContentState(
```
agatx, line 1358: **🟠 Outgoing content revision not seeded from snapshot on reconnect** (CONFIRMED)

Web added `MediaEngine.seedContentRevision` + `seedLocalContentRevisionFromSnapshot` to advance the outgoing content revision past the value persisted in `joined`/`room_state` after a reconnect. iOS (here) and Android (`SerenadaSession.kt:1937`) keep the outgoing counter only in memory and add no snapshot seeding.

If the router's `outgoingContentRevision` resets on reconnect (new router/session), the next `content_state` can carry a revision ≤ the one a same-sid peer already tracked. That peer's keep-highest gate discards it, so toggling the share off/on after reconnect is silently ignored on the remote and the screen-share indicator stays wrong. Port the snapshot seeding to iOS/Android.

agatx, line 1358: **🟠 Revisionless snapshot re-activates a stopped share (iOS)** (CONFIRMED)

`applyRemoteContentState` stores the gating revision as `content.revision` (no separate high-water field like web's `TrackedRemoteContent`), so a revisionless live update collapses the tracked revision to 0. The persisted `room_state` snapshot is then re-applied through this same reconciler on **every** `refreshRemoteParticipants` rebuild (~1517-1524, `isSnapshot:true`), and because snapshot revision N>0 > cached 0, it supersedes the stop.

Scenario: a remote peer stops a share via a revisionless `content_state{active:false}` while the persisted snapshot is still `{active:true, revision:N}`. On the next UI rebuild the cached stop (rev 0) is overwritten by the active snapshot → the stopped peer's indicator flips back **on** and stays on. Web doesn't exhibit this (separate tracked high-water keeps N). Keep a separate high-water revision on iOS as web does.

client/packages/core/src/SerenadaSession.ts
```diff
@@ -114,6 +139,12 @@ function toRoomParticipant(participant: SignalingProviderParticipant): RoomParti
...
         videoEnabled: participant.videoEnabled,
         connectionStatus: participant.connectionStatus,
+        capabilities: participant.capabilities,
+        mediaPolicy: participant.mediaPolicy,
+        // Carried through verbatim so the snapshot content state (a peer already
+        // sharing on join/reconnect) survives the provider boundary; the session
+        // reconciles it via the live keep-highest revision rule.
+        contentState: participant.contentState,
     };
 }
```
agatx, line 151: **🟠 Revisionless snapshot can overwrite a fresher live content state (web)** (CONFIRMED)

`applyRemoteContentState` has no snapshot-vs-live distinction for a *revisionless* update, unlike iOS which keeps the cache when a revisionless snapshot arrives (a live message is more authoritative). On web, a `room_state` snapshot for a peer whose `contentState` carries no revision is treated identically to a live message and overwrites the cached presentation state.

If a peer is sharing → live `content_state` arrives → a slightly stale revisionless `room_state` snapshot follows, the web viewer can revert the screen-share indicator to the snapshot's older `active`/`type` value, mis-showing share state. Add an `isSnapshot` path that keeps the live cache for revisionless snapshots, matching iOS. (Reconciler also at `SerenadaSession.ts:1265`.)

> _Anchored here because `SerenadaSession.ts:213` is outside the PR diff; the issue is at line 213._

client-android/serenada-core/src/main/java/app/serenada/core/call/SignalingMessageRouter.kt
```diff
@@ -134,7 +120,8 @@ internal class SignalingMessageRouter(
                 val fromCid = payload?.optString("from")?.ifBlank { null } ?: message.from
                 val active = payload?.optBoolean("active") ?: false
                 val contentType = if (active) payload?.optString("contentType")?.ifBlank { null } else null
-                onContentStateReceived(fromCid, active, contentType)
+                val revision = if (payload?.has("revision") == true) payload.optLong("revision") else null
```
agatx: **🟡 Malformed `revision` coerced to 0 on Android relay path** (PLAUSIBLE)

Live `content_state` revision is parsed with `payload.optLong("revision")`, which returns 0 for a present-but-non-numeric value — conflating a malformed revision with a valid 0. The server guard drops such values, but this peer-relay live path does not validate.

A `content_state` with `revision` present but non-integer (string/fractional from a non-conformant/forward client) is admitted as 0 and can wrongly seed the gate, potentially discarding the sender's genuine first `revision:1` and leaving the remote screen-share indicator unset for that toggle. Distinguish "absent" from "present but malformed" (e.g. `has("revision")` + typed check) before treating it as a tracked revision.

client-ios/SerenadaCore/Sources/Call/PeerConnectionSlot.swift
```diff
@@ -295,37 +341,242 @@ internal final class PeerConnectionSlot: PeerConnectionSlotProtocol {
...
+        onRenegotiationNeeded(remoteCid)
+    }
+
+    private func ensureRoleSendCapable(_ transceiver: RTCRtpTransceiver) {
+        if transceiver.direction != .sendRecv && !transceiver.isStopped {
+            transceiver.setDirection(.sendRecv, error: nil)
+        }
+    }
+
+    /// Conservative content (screen share) sender profile for the independent
```
agatx, line 534: **🟡 Content inbound bytes mis-attributed to camera role** (PLAUSIBLE)

`collectInboundRoleBytes` attributes every video inbound-rtp stat to the camera role whenever the content transceiver's receiver track id is empty/nil (not yet bound, or `trackIdentifier` not populated by libwebrtc). Content bytes get folded into `cameraBytes`, so `contentReceiving` can never go true even while the screen share is actively flowing.

A consumer reading `content.active && !contentReceiving` as a stall indicator then misreports a healthy remote screen share as stalled and shows a spurious "content not receiving" warning for the whole share. Consider falling back to mid/transceiver identity when the track id is unavailable.

client-ios/SerenadaCore/Sources/Call/PeerConnectionSlot.swift
```diff
@@ -347,6 +598,79 @@ internal final class PeerConnectionSlot: PeerConnectionSlotProtocol {
...
+        }
+        // No transceiver (legacy didAdd:stream delivery). For a capable peer this
+        // path cannot distinguish camera from content, and because native libwebrtc
+        // hands each callback a DISTINCT RTCVideoTrack wrapper for the same track, it
+        // re-set (clobbered) the already-classified remote camera with a churned or
+        // content wrapper on every stream callback -> remote video blanked on iOS.
+        // The transceiver-carrying didAdd:rtpReceiver / didStartReceivingOn paths are
+        // authoritative for capable peers, so ignore the nil-transceiver delivery.
+        // (Android's onAddStream is a no-op, which is why this only bit iOS.)
+        logger?.log(.debug, tag: "PeerConnection", "[\(remoteCid)] Ignoring nil-transceiver remote video delivery (capable peer)")
```
agatx: **🟡 Stale remote camera track on capable-peer stream removal** (PLAUSIBLE)

For a capable peer, `didRemove stream` calls `routeRemoteVideoTrack(nil, nil)`; the capable-peer nil-transceiver branch logs and returns **without** clearing `remoteVideoTrack`, so the remote camera track is never torn down on a Plan-B-style stream removal.

If a remote independent-capable peer's media is removed via the legacy `didRemove:stream` delivery (no transceiver), iOS ignores it and leaves the stale `remoteVideoTrack` bound to renderers → the viewer keeps seeing the peer's last (frozen) camera frame instead of it clearing. Legacy peers clear correctly; only capable peers are affected. Clear `remoteVideoTrack` in the capable-peer stream-removal branch.

client-ios/SerenadaCore/Sources/SerenadaSession.swift
```diff
@@ -654,30 +695,116 @@ public final class SerenadaSession: ObservableObject {
     /// Start screen sharing via the broadcast upload extension.
```
agatx, line 706: **🟡 Double `content_state` emit causes remote flicker on independent stop** (CONFIRMED)

When the local camera is in world/composite mode, stopping an independent screen share fires two back-to-back `content_state` messages: `onScreenShareStopped` → `broadcastLocalContentState(active:false)` (rev N), immediately followed by `restoreCameraContentHintAfterIndependentStop()` (~line 224) → `broadcastLocalContentState(active:true, contentType: world/composite)` (rev N+1).

Both are delivered/ordered fine, but remote peers see content go inactive then immediately active again in the same tick → a visible flicker of the remote content tile that the legacy single-`active:false` stop path didn't produce. This is the documented intent (restore the suppressed camera-framing hint) but worth a reviewer's eye — consider coalescing into one emit.

> _Anchored here because `SerenadaSession.swift:647` is outside the PR diff; the issue is at line 647._

client/packages/core/src/media/MediaEngine.ts
```diff
@@ -462,13 +639,272 @@ export class MediaEngine {
...
+                // Legacy peer: swap the single video sender's track to the display track.
+                if (await this.swapLegacyPeerVideoTrack(remoteCid, peer, displayTrack)) {
+                    attachedCount += 1;
+                }
+            }
+        }));
+
+        // 3) Roll back only if the share can never flow anywhere.
+        if (attachedCount === 0 && pendingCount === 0 && this.peers.size > 0) {
+            this.sendContentState({ active: false });
```
agatx: **🟡 False "screen share started" flicker on full attach failure (web)** (PLAUSIBLE)

`startScreenShareIndependent` broadcasts `content_state{active:true}` before per-peer attach, then on full-attach-failure broadcasts `{active:false}`. When no peer can attach (e.g. every capable peer's `replaceTrack` rejects and none is left pending, with peers present), remote participants first see `active:true` (UI shows "X is sharing" / opens a content tile), then `active:false` a moment later — a brief false indicator that flickers off with no content ever rendered. Consider only broadcasting `active:true` after at least one successful attach.

client-ios/SerenadaCore/Sources/SerenadaSession.swift
```diff
@@ -789,6 +916,19 @@ public final class SerenadaSession: ObservableObject {
...
+    /// Detach a previously attached remote content renderer for a peer.
+    public func detachRemoteContentRenderer(_ renderer: AnyObject, forParticipant cid: String) { peerSlots[cid]?.detachRemoteContentRenderer(renderer) }
+    /// Attach a renderer to the LOCAL content (screen share) track for local preview.
+    public func attachLocalContentRenderer(_ renderer: AnyObject) { webRtcEngine.attachLocalContentRenderer(renderer) }
+    /// Detach a previously attached local content renderer.
+    public func detachLocalContentRenderer(_ renderer: AnyObject) { webRtcEngine.detachLocalContentRenderer(renderer) }
+
     // MARK: - Join Flow
 
     private func beginJoinIfNeeded() async {
```
agatx, line 1046: **🔵 Cleanup — redundant single-pointer content diagnostics** (CONFIRMED)

iOS and Android maintain both the legacy single-pointer content diagnostics (`remoteContentParticipantId`/`remoteContentType`) **and** the new per-participant content map, reconciling the redundant pointer with per-platform-divergent ad hoc logic (iOS inline ternary; Android a dedicated `refreshContentDiagnosticsPointer` helper) while web dropped the pointer entirely.

The same concept ("which remote is presenting") is modeled at two altitudes; the pointer must be hand-reconciled on every `content_state`, peer-leave, and snapshot path, three different ways across platforms — so a multi-sharer/rapid-toggle edge case can desync the pointer on one platform but not another. The per-participant map already fully determines this value; consider deriving the pointer or dropping it as web did.

> _Anchored here because `SerenadaSession.swift:1018` is outside the PR diff; the issue is at line 1018._

client-ios/SerenadaCore/Sources/SerenadaSession.swift
```diff
@@ -1032,12 +1174,110 @@ public final class SerenadaSession: ObservableObject {
...
+        sid: String?,
+        active: Bool,
+        type: String?,
+        revision: Int64?,
+        isSnapshot: Bool
+    ) -> ParticipantContent? {
+        let existing = remoteContentStates[cid]
+
+        if let incomingRevision = revision, let existing {
+            let sameSession = sidsMatch(existing.sid, sid)
```
agatx, line 1369: **🔵 Cleanup — inert `(cid, sid)` revision-scoping infrastructure** (CONFIRMED)

iOS (here) and web (`SerenadaSession.ts:1209`) layer a `(cid, sid)` revision-scoping mechanism (`sidsMatch`, the `(sid, content)` tuple, the `sameSession` branch) on top of a relay that **never carries a sender sid**. `signaling.go relayMsg` sets only payload `from`, never the envelope SID, so `message.sid` is always nil for `content_state`. `sidsMatch(nil,nil)` is always true and `existing.sid === sid` is always true, so the `sameSession` false path can never execute.

The result is dead infrastructure that two platforms maintain (extra fields, helpers, branches, tests) while Android took the honest cid-only route — three SDKs now model the same wire contract at two depths, and a future reader will assume sid-scoping works. Drop the inert sid path or actually thread the sender sid through the relay.

client/packages/core/src/SerenadaSession.ts
```diff
@@ -1362,8 +1620,22 @@ export class SerenadaSession implements SerenadaSessionHandle {
         if (!this.isConnected || this.roomState === null) return;
         this.mediaLivenessEmitInFlight = true;
         try {
+            // Refresh per-role inbound liveness on the same cadence and re-render
+            // so `cameraReceiving`/`contentReceiving` on remote participants stay
+            // current for stall diagnostics. Independent of the server-facing
+            // `media_liveness` emit below (which is gated on connection state).
+            await this.media.sampleInboundRoleLiveness();
```
agatx: **🔵 Cleanup — `getStats()` collected twice per peer per tick** (CONFIRMED)

`emitMediaLiveness` gathers WebRTC `getStats()` twice per peer per tick and serially: `sampleInboundRoleLiveness()` iterates `this.peers` calling `getStats()`, then `getInboundFlowingCids()` iterates the same peers calling `getStats()` again. `getStats()` is the expensive call here, and the two `await`s are sequential, so the tick's main-thread latency is also doubled — for the lifetime of the call.

Both passes need the same report (one sums all inbound-rtp bytes for the server `media_liveness` signal, the other splits video bytes by camera/content role). Collect once per peer per tick and derive both. Same pattern on Android (`SerenadaSession.kt:2041`), iOS (`PeerConnectionSlot.swift:972`), and `MediaEngine.ts:660`.

client-ios/SerenadaCore/Sources/SerenadaConfig.swift
agatx: Fixed in 2db5e860. The bundled iOS host now explicitly enables independent content video and passes the ReplayKit broadcast IPC config (`enableIndependentContentVideo: true`, `.broadcast(...)` with the app group and broadcast extension bundle id). The SDK default remains disabled for integrators unless they opt in; the broadcast-extension doc snippet and README wording were updated to make that explicit.

client-ios/SerenadaCore/Sources/Call/ScreenShareCapturers.swift
agatx: Fixed in 2db5e860. `BroadcastFrameReader.pollFrame` now validates the advertised plane layout against the mapped shared-memory size before allocation/copy, rejects unsupported plane counts/overflowing byte counts, validates destination plane heights, and re-checks the seqlock before copying so a writer update cannot race the sampled header.

client-ios/SerenadaCore/Sources/SerenadaSession.swift
agatx: Fixed in 2db5e860. iOS and Android now emit `media_liveness` with an empty `cids` array while connected and seeing no inbound flow, matching the web behavior. Added/updated platform tests for the empty heartbeat case.

server/signaling.go
agatx: Fixed in 2db5e860. Server reconnect handling now preserves omitted `capabilities` / `mediaPolicy` records, and when `capabilities` is present it merges only present allowlisted keys so a reconnect carrying `maxParticipants` does not erase `independentContentVideo`. A present empty object still clears back to defaults. Added a server regression test.

client-ios/SerenadaCore/Sources/SerenadaSession.swift
agatx: Fixed in 2db5e860. iOS and Android now seed the outgoing local content-state revision from the recovered local participant snapshot before broadcasting again, matching the web recovery behavior. Added iOS and Android regression tests.

agatx: Fixed in 2db5e860. iOS remote content tracking now keeps a separate high-water revision from the public `ParticipantContent.revision`, so a revisionless live stop no longer lowers the revision gate and an older snapshot cannot reactivate the share. Added the revisionless-stop regression test.

client/packages/core/src/SerenadaSession.ts
agatx: Fixed in 2db5e860. Web `applyRemoteContentState` now distinguishes snapshot vs live updates; a revisionless snapshot keeps an existing live cache instead of overwriting fresher state. Added a web regression test.

client-android/serenada-core/src/main/java/app/serenada/core/call/SignalingMessageRouter.kt
agatx: Fixed in 2db5e860. Android content-state revision parsing now treats malformed, string, fractional, negative, and overflowing revisions as absent instead of coercing them to `0`; the relay path uses the same parser. Added payload/parser and high-water regression tests.

client-ios/SerenadaCore/Sources/Call/PeerConnectionSlot.swift
agatx: Fixed in 2db5e860. iOS inbound role-byte attribution now falls back to the content transceiver `mid` when the receiver track id is unavailable, so content bytes are not miscounted as camera bytes during unbound/recreated receiver windows.

client-ios/SerenadaCore/Sources/Call/PeerConnectionSlot.swift
agatx: Fixed in 2db5e860. The capable-peer nil-transceiver removal path now clears the remote camera track instead of logging and returning, so `didRemove stream` cannot leave a stale camera track published.

client-ios/SerenadaCore/Sources/SerenadaSession.swift
agatx: Fixed in 2db5e860. Mobile independent-share stop now emits a single final `content_state`: active world/composite camera content when that camera framing remains, otherwise inactive. This removes the inactive-then-active flicker. Android contract expectations were updated too.

client/packages/core/src/media/MediaEngine.ts
agatx: Fixed in 2db5e860. Web independent screen share now delays the active `content_state` until at least one attach/pending path succeeds; a full attach failure releases capture without broadcasting a transient active/inactive pair. Updated the failure test.

client-ios/SerenadaCore/Sources/SerenadaSession.swift
agatx: Fixed in 2db5e860. iOS no longer updates the legacy single diagnostics pointer independently in the content-state handler; it derives the pointer from the per-CID content-state map during participant refresh. Android was already derived, and web does not keep the same single-pointer mirror.

client-ios/SerenadaCore/Sources/SerenadaSession.swift
agatx: Resolved by clarification in 2db5e860. I kept the `(cid, sid)` revision-scoping path intentionally: hosted signaling currently omits sender `sid`, but custom `SignalingProvider` implementations can surface it, and the public contract documents new-session supersede semantics. Comments now state that hosted traffic follows the cid-keyed path while custom providers get sid-aware behavior.

client/packages/core/src/SerenadaSession.ts
agatx: Fixed in 2db5e860. Web now samples total inbound flow and per-role video liveness from one `getStats()` report per peer via `sampleInboundLiveness()`, and `SerenadaSession` uses that combined sampler for the media-liveness tick. Added a test asserting one `getStats()` call per peer while refreshing both outputs.

agatx: Follow-up in 2b072282 fully closes the mobile side too. iOS and Android now collect total inbound flow plus role-split video liveness from one stats report per peer per tick, and the mobile tests assert the timer no longer calls the old bytes-only or role-only collectors.

PR #137, opened 2026-06-27
Multi-call session: hold + switch between calls (SDK + host apps + server)

agatx:
## Summary

Implements the v1 multi-call session design (`docs/multi-call-session-design.md`): a host app can keep multiple Serenada calls in progress and switch the single foreground media owner between them. Held calls stay joined and signaled but own no microphone, camera, screen share, or audible playout. **Existing single-call APIs are unchanged** — single-call is just a registry with one foreground call.

Built across all three SDKs **and the Go server**, in 6 phases. Each phase ran the same loop: implement per platform → codex review → simplify → codex verify → reconcile cross-platform → commit. Every phase converged to a codex **PASS**.

Authoritative spec: `docs/multi-call-session-design.md` + the derived cross-platform contract `docs/multi-call-session-contract.md`. Integration guide: `docs/multi-call-session.md`.

## What's in it

- **Process-wide foreground media arbiter** — one lease per process, owner-token + operation-generation double-fence, owning-mode (direct vs registry).
- **`SerenadaCallRegistry`** — `joinHeld` / `joinAndSwitch` / `switchTo` / `hold` / `leave` / `end`; serialized op queue; rollback switch; call-identity dedup; aggregate state.
- **Session hold/resume** — real mic + camera capture release (not just `enabled=false`), remote-playout deafen, stable send-capable held senders so resume attaches via `replaceTrack` with no renegotiation.
- **`held` peer-signaling** — additive field on `participant_media_state` (+ server relay/snapshot, see below).
- **Host apps stay single-call** (productization decision, `aa1c0a37`): the registry-backed switcher UX lives in the purpose-built multi-call **samples**; the bundled iOS/Android `CallManager`s only gained a teardown-before-switch so deep-link room switching works under the new process-wide lease. The web `useSerenadaCallRegistry` hook ships in `react-ui`.
- **Audio-coordinator lease/generation fencing** (iOS/Android) so a superseded session can't re-drive OS audio/focus/route after a switch.
- **Docs + a minimal multi-call sample per platform** (web sample host is configurable via `?host=`).

## Server change worth a look

The design assumed peer messages relay verbatim. The server actually **allowlists** `participant_media_state` fields, so it was silently dropping `held` — the feature would have failed end-to-end on a real server while all client unit tests (fakes relay verbatim) passed. Fixed: the server now parses, relays, and snapshots `held` (additive, `*bool`/omitempty, backward-compatible). Caught by codex reading the Go server beyond the diff.

## Tests / status

- Web **678**, iOS **629 + app-bundle suites** (TEST SUCCEEDED), Android **590** across `:serenada-core` + `:app`, Server `go test ./...` green (incl. new held-state tests). Counts as of the ship-readiness review round.
- Parity: resilience constants **30/30**, version + signaling-protocol OK.
- The double-codex gate caught **~50 real defects** across phases — mostly cross-platform consistency gaps (one platform right, others wrong) plus concurrency/lifecycle bugs.
- **Browser-validated**: dogfooded the web sample against a local branch server — registry, `joinHeld`, preflight `needsPermission` (Core Invariant 4), held-calls surface, and `leave` all work in-browser.

## Remaining: on-device manual validation (not unit-testable)

Needs two real devices: mic/camera indicators follow only the foreground call; held remote audio silent; Bluetooth route survives repeated switches; Android FGS type follows the active call (service stays up while any call exists; no auto-promote); two-peer `held` propagation; screen share stops cleanly on switch; failed activation restores the previous call.

## Not in this PR

`VERSION` is **not** bumped (release/ship is a separate step). Multi-record recovery and a full prebuilt multi-call switcher UI are fast-follows per the design. V1 recovery stays foreground-only.

History: ~30 commits, one per phase/round (`feat` impl → per-codex-round `fix` → `refactor` simplify → server fix).

https://claude.ai/code/session_01PR5mAquTznSeMrWvEsNRT9

## Ship-readiness review round (2026-07-01/02)

A multi-model review pass (5 specialist agents + Claude adversarial + Codex) found and fixed two P1 lease-wedge bugs (`67bc9dac`) and three P2s (`b37175dd`), plus doc accuracy (`803688f6`):

- **P1** Android registry `close()` leaked the process-wide foreground lease (and had no `closed` guard) — wedged every later join until process restart; reachable from the sample's "Close Registry" button.
- **P1** All platforms: `joinAndSwitch`/`switchTo` on a reused *still-joining* call switched before membership settled; a join timeout after that switch stranded the lease (the failure handlers assumed held joins never own it). Switch entry points now gate on join settlement.
- **P2** Android resume never re-acquired the CPU wake lock (screen-off audio could die); iOS held calls kept 10Hz/2Hz pollers running (battery); stale relay cache beat the fresh reconnect snapshot for remote `held` on all platforms.
- Deferred (documented, non-blocking): drain-timeout recovery gap on the no-op switch path, `switchTo(endedCall)` disturbing the active call (web/Android), iOS `preconditionFailure` on a host-less bare-roomId RoomRef, held-call media-liveness eviction question, error-taxonomy/retention cross-platform convergence, web `joinInternal` visible in published `.d.ts` (no `stripInternal`).


client/packages/react-ui/src/hooks/useSerenadaCallRegistry.ts
```diff
@@ -0,0 +1,174 @@
...
+        // plus the registry-specific retention. A reconstruct tears down the old
+        // registry (cleanup effect below) and builds a fresh one.
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+        [
+            config.serverHost,
+            config.defaultAudioEnabled,
+            config.defaultVideoEnabled,
+            transportsKey,
+            config.turnsOnly,
+            config.logger,
```
agatx: **P1 — `config.logger` in the registry `useMemo` deps silently tears down every live call**

`useSerenadaSession` deliberately keys its session `useMemo` on only the primitive config fields (`serverHost`, `defaultAudioEnabled`, `defaultVideoEnabled`, `transportsKey`, `turnsOnly`) and omits `logger`. This array adds `config.logger`, which is a function. A host passing an inline logger (`logger: { log: (...) => console.log(...) }`, the common React idiom) mints a new logger identity on every parent re-render, so the deps change, the registry is reconstructed, and the `[registry]`-keyed cleanup effect (line 126) fires `registry.leave(call.id)` for every managed call.

Net effect: the active foreground call **and all held calls are torn down mid-session on an unrelated re-render** and the user is silently dropped. The comment on line 109 ("the same config keys useSerenadaSession keys on") is inaccurate — `useSerenadaSession` does not key on `logger`.

Removing this line restores parity; the logger is still captured by closure (line 106), and a stale logger after a logger swap is the same accepted tradeoff `useSerenadaSession` already makes.

```suggestion
```

client/packages/core/src/SerenadaSession.ts
```diff
@@ -1121,51 +1324,398 @@ export class SerenadaSession implements SerenadaSessionHandle {
...
+        if (this.mediaRole === 'foreground') return;
+        const { gen, token, throwOnFailure } = opts;
+        this.resumeInFlight = true;
+        // Surface activation progress (contract §4; parity with iOS/Android):
+        // `activating` while reacquiring media, `active` on success, `inactive`
+        // on a superseded/failed resume (the rollback path).
+        this.mediaActivationState = 'activating';
+        this.rebuildState();
+        try {
+            await this.media.resumeLocalMediaFromHold(this.desiredAudioEnabled, this.desiredVideoMode);
```
agatx, line 1580: **P1 — Web resume from hold restores the wrong camera facing (parity break vs iOS/Android)**

`resumeLocalMediaFromHold` is handed `desiredVideoMode`, but `MediaEngine.reacquireVideoTrack` only treats it as a boolean (`if (desiredVideoMode !== 'off')`) and reacquires with the stale `this.facingMode`. Per Core Invariant 2, `setCameraMode`/`flipCamera` while held only updates `desiredVideoMode` and defers the actual switch to resume — but resume never reconciles to it.

Repro: hold a call → switch camera to world/back while held → resume. The camera comes back on the **old facing** (front/selfie) for both the local user and the peer. iOS and Android apply the specific mode on resume (Android calls `restartVideoCapturerWithFallback(cameraSourceFromMode(...))`), so this is web-only.

Fix: have `reacquireVideoTrack` (or `resumeLocalMediaFromHold`) reconcile `this.facingMode` to `desiredVideoMode` before reacquiring, instead of checking only `!== 'off'`.

client/packages/core/src/SerenadaSession.ts
```diff
@@ -1121,51 +1324,398 @@ export class SerenadaSession implements SerenadaSessionHandle {
...
+            if (this.isInactive) return;
+            // Both fences: a bumped generation (a concurrent hold) OR a lost lease
+            // (token no longer the arbiter's current owner) supersedes this resume.
+            const supersededByGeneration = this.mediaOpGeneration !== gen;
+            const supersededByToken = token !== null && !foregroundArbiter.isCurrentOwner(token);
+            if (supersededByGeneration || supersededByToken) {
```
agatx, line 1586: **P2 — `runResume` superseded branch resolves without throwing on the token-gated path (latent)**

The superseded branch (`supersededByGeneration || supersededByToken`) calls `rollbackSupersededResume()` and returns **without throwing even when `throwOnFailure` is true** — the `activateForeground` caller. But the doc comments on `activateForeground` (~line 1614) and `runResume` (~1488-1490) both promise it "re-throws so the registry switch can roll back."

`SerenadaCallRegistry.runSwitch` (`SerenadaCallRegistry.ts:510`) awaits `activateForeground` and, on a normal resolution, treats activation as succeeded: sets `activeCallId = next.id`, marks active, publishes, returns `{kind:'active'}`. So if a resume is superseded mid-reacquire, the session rolls `next` back to **HELD** (no `held:false` broadcast) while the registry believes `next` is the live foreground call, and the arbiter still holds `next`'s lease.

iOS does not have this hole: `completeForegroundActivation`/`failForegroundActivation` leave `mediaActivationState != .active` and the registry polls that state, so a superseded resume can't read as success. Today the web path is gated by the registry's promise-chain serialization (nothing bumps the generation during the awaited activation), so it's latent — but it goes live the moment any un-queued generation bump or lease rotation can interleave. Recommend throwing on the superseded branch when `throwOnFailure` is true, matching the documented contract.

client/packages/core/src/media/MediaEngine.ts
```diff
@@ -541,6 +562,181 @@ export class MediaEngine {
...
+        // exists, so this is a no-op (single-call behavior unchanged).
+        if (!this.localStream) this.localStream = new MediaStream();
+        if (this.localStream.getAudioTracks()[0]) return;
+        if (this.audioRecoveryInFlight || this.requestingMedia) return;
+        this.audioRecoveryInFlight = true;
+        try {
+            const devices = await this.enumerateMediaDevices();
+            const preferredInput = this.selectPreferredAudioInput(devices, null);
+            const track = await this.acquireAudioTrack(true, preferredInput.device?.deviceId);
+            if (this.destroyed || !this.localStream) {
```
agatx: **P2 — `reacquireLocalAudioCapture` doesn't re-check `heldNoCapture` after the await; a hold racing a resume can briefly publish live mic to peers**

After `await this.acquireAudioTrack(...)`, the post-await guard re-checks only `this.destroyed` and `this.localStream` — not `this.heldNoCapture`. During `resumeLocalMediaFromHold`, `heldNoCapture` is cleared and `getUserMedia` is awaited; if a concurrent `suspendForHold` re-latches `heldNoCapture=true` and releases capture during that await, the resolved track is still `addTrack`'d and `replaceTrack`'d onto every peer's audio sender.

The session-level supersession fence re-runs `suspendLocalMediaForHold` to release it, so the leak self-heals — but for the window between attach and the fence, a held call publishes live mic audio to peers, which a held call must never do (Core Invariant 2 / privacy). Add a `this.heldNoCapture` check to the post-await guard so the freshly captured track is dropped instead of attached.

client/packages/core/src/media/MediaEngine.ts
```diff
@@ -541,6 +562,181 @@ export class MediaEngine {
...
+    async reacquireLocalAudioCapture(): Promise<void> {
+        // Held-initial resume (Core Invariant 3 / contract §5): a session joined
+        // directly into `held` never ran `startLocalMedia`, so `localStream` is
+        // null here. Create an empty stream so the freshly acquired mic can be
+        // added and attached to the existing held audio senders via
+        // `replaceTrack`. For a normally-joined call `localStream` already
+        // exists, so this is a no-op (single-call behavior unchanged).
+        if (!this.localStream) this.localStream = new MediaStream();
+        if (this.localStream.getAudioTracks()[0]) return;
+        if (this.audioRecoveryInFlight || this.requestingMedia) return;
```
agatx: **P2 — unmute during an in-flight audio reacquire silently no-ops and broadcasts `audioEnabled:false`**

`reacquireLocalAudioCapture()` returns early (no track acquired) when `audioRecoveryInFlight || requestingMedia`. The `SerenadaSession` unmute caller then derives `actualAudioPublished` from the still-null track and broadcasts `participant_media_state { held:false, audioEnabled:false }`.

Repro: user taps unmute while another audio acquire is in flight (e.g. a `resumeForeground`/`resumeLocalMediaFromHold` audio acquire overlapping a manual `toggleAudio(true)` on a call that resumed muted). The local user believes they unmuted, but the peer sees them muted and hears nothing, with no follow-up event to correct it until the next toggle. Consider queuing/retrying the unmute after the in-flight acquire settles, or surfacing the no-op so the caller doesn't broadcast a false `audioEnabled:false`.

client-android/serenada-core/src/main/java/app/serenada/core/SerenadaCallRegistry.kt
```diff
@@ -0,0 +1,727 @@
...
+    private fun ManagedCall.toManagedCallState(): ManagedCallState {
+        val s = session.state.value
+        // The role is the registry's authoritative view (lease ownership), NOT the
+        // session's mutable `mediaRole`: a session-internal teardown resets the
+        // latter to FOREGROUND, which would mislabel an ended/held call.
+        val role = if (!ended && foregroundToken != null) CallMediaRole.FOREGROUND else CallMediaRole.HELD
+        // Held/ended calls sit at INACTIVE; the live session value is only
+        // meaningful while the call is the foreground one being activated. Keep a
+        // surfaced NEEDS_PERMISSION/FAILED (set on the call without a transition).
+        val activation = when {
```
agatx: **P3 — Android alone re-derives `mediaActivationState` for held calls in the snapshot; diverges published `ManagedCallState` from web/iOS**

This `when` block re-derives `NEEDS_PERMISSION`/`FAILED` for held calls in the published snapshot, and the registry also mutates the held session's `mediaActivationState` (via `setMediaActivationState` in the switch/hold-timeout path) **and** surfaces the same signal on the orthogonal `activationError` field. Web and iOS keep `mediaActivationState` a pure function of the session's own foreground lifecycle (held == `INACTIVE`, per the `MediaActivationState` doc) and carry permission/failure once, on `activationError`.

Result: a host reading `mediaActivationState` to drive UI sees `NEEDS_PERMISSION` on Android but `INACTIVE` on web/iOS for the identical scenario (a held call needing permission). The Android-only set-site and this snapshot `when` must be kept in sync by hand or the held-call value silently drifts. Recommend converging on the web/iOS contract: route permission/failure solely through `activationError` and drop the special-case.

client-ios/SerenadaCore/Sources/SerenadaCallRegistry.swift
```diff
@@ -0,0 +1,1033 @@
...
+            }
+            for _ in 0..<4 { await Task.yield() }
+            if condition() { return true }
+        }
+        return condition()
+    }
+
+    // MARK: - Publishing
+
+    private func publish() async {
```
agatx, line 1245: **P3 — iOS under-publishes held-call field changes; web/Android over-publish — three SDKs, three different contracts**

iOS only observes `registryPhasePublisher` (terminal phases) per call and calls `publish()` at the operation-boundary sites, so a held call's `participantCount`, `membershipPhase` (e.g. `waiting`→`inCall`), or `actualAudioPublished` changes are **not** reflected in `registry.calls` until the next registry operation. The same multi-call switcher UI updates live on web/Android but shows a stale held-call chip on iOS until the user performs a switch/hold/leave.

Web and Android have the opposite problem (they re-publish on every stats tick — see the `publish()`-per-emission finding). The three platforms should converge on one contract: forward the held-call fields the host renders, **coalesced** rather than per-stats-tick.

client/packages/core/src/SerenadaCallRegistry.ts
```diff
@@ -0,0 +1,889 @@
...
+        // hold for ANY managed call — active OR held — not just the active one:
+        // a held call that ends on its own would otherwise wedge registry mode
+        // forever, and an active call that ends would keep the lease through the
+        // ~3s ending screen. `'ending'` is treated as terminal for lease purposes
+        // (release promptly — don't wait through the ending screen). Routed
+        // through the SAME operation queue so it cannot interleave a switch.
+        if (this.isSessionTerminal(call) && !call.terminated) {
+            void this.enqueue(() => this.handleCallTerminated(call));
+        }
+        this.publish();
```
agatx, line 750: **P3 — `publish()` on every session emission recomputes the full N-call snapshot on each held-call stats tick (O(N²))**

`onSessionStateChange` calls `publish()` on every session emission. Once a held call reaches phase `inCall` it starts periodic stats collection, which fires `notifyListeners()` on every sample (`SerenadaSession.ts:2305`). Each tick routes into `publish()`, which maps **every** managed call through `toManagedCallState()` (≈14 session getters each) and fans out to all registry listeners.

With N held calls each ticking, that's O(N²) recompute + notify per round of stats churn, plus a React re-render of the whole switcher on every held call's stats sample — even though the foreground renderer ignores held-call stats. Cheaper: publish only the changed call's snapshot, debounce/coalesce held-call ticks, or skip `publish()` for stats-only emissions of non-active calls. (Pairs with the iOS under-publish finding — pick one coalesced contract for all three.)

client/packages/react-ui/src/hooks/useSerenadaCallRegistry.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    const state = useSyncExternalStore(
+        useMemo(() => (onChange: () => void) => registry.subscribe(onChange), [registry]),
+        () => registry.state,
+        () => EMPTY_REGISTRY_STATE,
+    );
+
+    // The active call carries the live session, so it is read off the registry
+    // (not the value snapshot). Re-derive on each state change.
+    const activeCall = registry.activeCall;
```
agatx: **P3 — `registry.activeCall` mints a fresh `{state, session}` object every render**

`const activeCall = registry.activeCall` calls the getter, which constructs a new `{ state, session }` wrapper on every render (`SerenadaCallRegistry.ts:174`). Its identity changes every render, so any consumer that puts `activeCall` in a `useEffect`/`useMemo` dep array — or a memoized child receiving it as a prop — re-runs/re-renders on every registry state change even when the underlying session and snapshot are unchanged. Wrap it in a `useMemo` keyed on `activeCallId` + the active call's snapshot identity so the object identity is stable across unrelated churn.

client-android/serenada-core/src/main/java/app/serenada/core/CallRegistryTypes.kt
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * The canonical room id used by [SerenadaCallRegistry] for duplicate-join dedup
+ * (contract §7 "Call identity policy"). Host-agnostic: a `/call/<token>` URL
+ * collapses to `<token>` regardless of scheme/host/query/fragment/trailing slash,
+ * so `serenada.app` and `serenada-app.ru` URLs for the same room are one key; a
+ * bare id is taken verbatim. Reuses the same last-path-segment extraction the
+ * single-call join already uses (`SerenadaCore.resolveRoomUrl`); no new format.
+ */
+internal fun canonicalRoomId(room: RoomRef): String = when (room) {
```
agatx, line 42: **P3 — `canonicalRoomId()` re-implements room-token extraction instead of sharing `resolveRoomUrl`'s logic**

This copies the `Uri.parse(...).lastPathSegment` + `substringAfterLast('/')` fallback that `SerenadaCore.resolveRoomUrl` already does, making Android the only platform with a third divergent copy of room-token parsing — iOS factored this into `DeepLinkParser.extractRoomId` and web into `roomIdentity.canonicalizeRoomId`. The doc comment says it "reuses" `resolveRoomUrl`, but it duplicates it.

If either parser is later tweaked (query/fragment/trailing-slash handling), the registry's dedup key can disagree with the room the join path actually connects to, so a duplicate join either fails to dedupe (two managed calls for one room) or wrongly dedupes. Fix: extract a shared `canonicalizeRoomId`/`extractRoomToken` and have both `resolveRoomUrl` and `canonicalRoomId` call it.

client-android/serenada-core/src/main/java/app/serenada/core/SerenadaCallRegistry.kt
```diff
@@ -0,0 +1,727 @@
...
+    fun session(callId: CallId): SerenadaSession? {
+        assertMainThread()
+        return managedCalls[callId]?.session
+    }
+
+    // The active call is the one holding the foreground lease — the registry's
+    // OWN authoritative signal ([ManagedCall.foregroundToken]), NOT the session's
+    // mutable `mediaRole` (which a session-internal teardown resets to FOREGROUND).
+    private val activeCallId: CallId?
+        get() = managedCalls.values.firstOrNull { !it.ended && it.foregroundToken != null }?.callId
```
agatx: **P4 — `activeCallId` linear-scans `managedCalls` on every access, including inside every `publish()`**

`activeCallId` scans `managedCalls.values.firstOrNull { !it.ended && it.foregroundToken != null }` on each read, and `publish()` reads it on every state recomputation. Combined with `publish()` firing on every session emission, the foreground lookup is re-scanned across all N calls on every tick of every session. Minor for small N, but it compounds the publish-per-tick churn. Cheaper: maintain `activeCallId` as a stored field updated at the lease acquire/release transitions, as iOS does.

client-ios/SerenadaCore/Sources/Call/SerenadaAudioCoordinator.swift
```diff
@@ -147,6 +147,119 @@ public enum AudioCoordinatorEvent: Sendable {
...
+    /// IMMEDIATELY before `activateCallSession` (Phase 4, contract §6). Installing
+    /// it any earlier lets a still-pending old `deactivateCallSession` (running as
+    /// the previous task drains) consume/clear the newly-installed lease before the
+    /// matching activate runs. Records the lease both locally and in the
+    /// process-global ``AudioSessionLeaseRegistry`` so a cross-instance stale
+    /// callback can be fenced.
+    func setForegroundLease(_ lease: AudioSessionLease)
+    /// Clear the lease when the session deactivates. After this, any late callback
+    /// captured under the cleared lease is stale and dropped.
+    func clearForegroundLease(_ lease: AudioSessionLease)
```
agatx: **P4 — dead protocol method `clearForegroundLease(_:)` is never called**

`clearForegroundLease(_:)` is declared on `LeaseAwareAudioCoordinator` and implemented in `DefaultAudioCoordinator` (line 168), but a repo-wide search finds no call site in the SDK, host, or tests. The session's deactivation path fences via `installedLeaseSnapshot()` + `deactivateCallSession(fencedBy:)`, which clears the lease through the private `clearInstalledLease()`. The method's own doc ("Called by the session when the session deactivates") describes a path that never runs. Drop it from both the protocol and `DefaultAudioCoordinator` — removing it has zero behavioral effect and removes a misleading contract.

client/packages/core/src/SerenadaCallRegistry.ts
```diff
@@ -0,0 +1,889 @@
...
+        // (idempotency: lease already released, activeCallId already cleared).
+        call.terminated = true;
+        try {
+            if (kind === 'end') call.session.end();
+            else call.session.leave();
+        } catch (err) {
+            this.log('warning', `${kind} teardown failed: ${formatError(err)}`);
+        }
+        this.scheduleEndedRemoval(call);
+        this.releaseModeIfIdle();
```
agatx, line 720: **P4 — trailing `releaseModeIfIdle()` after `scheduleEndedRemoval()` reads as dead code but is load-bearing only on the retention path**

In `runLeaveOrEnd` (and identically in `handleCallTerminated`, ~690-694), `scheduleEndedRemoval(call)` is immediately followed by an explicit `releaseModeIfIdle()`. On the default path (`endedCallRetentionMs <= 0`), `scheduleEndedRemoval` → `removeCall` already calls `releaseModeIfIdle()` (line 731), so the trailing call is a redundant no-op. It only matters when `retention > 0` (the ended call lingers, so the mode must be freed separately) — which isn't obvious here.

A reader can't tell whether it's dead or load-bearing without tracing both branches of `scheduleEndedRemoval`; if they "clean it up," registry-mode release breaks on the retention>0 path (a later `SerenadaCore.join()` then fails with `ForegroundLeaseUnavailable`). Add a one-line comment noting it's only needed for the retention>0 case, or push the release into `scheduleEndedRemoval`.

client-ios/SerenadaCore/Sources/SerenadaCallRegistry.swift
```diff
@@ -0,0 +1,1033 @@
...
+    /// 1. A generous **yield-only** budget first: a normal settle (coordinator
+    ///    completion, join handlers) needs only a handful of main-actor hops, so
+    ///    this resolves the common path WITHOUT ever calling `clock.sleep`. This
+    ///    matters because the test fake clock only resolves a `sleep` on an
+    ///    explicit `advance`, so a sleep on the success path would deadlock.
+    /// 2. A **sleep + deadline** phase for a genuinely stuck operation (a blocked
+    ///    coordinator). A live clock trips the deadline on real elapsed time; a
+    ///    fake clock trips it when the test drives `advance(byMs:)` past the
+    ///    timeout. Returns whether the condition was met before the deadline.
+    private func awaitSettle(timeoutMs: Int, condition: @MainActor () -> Bool) async -> Bool {
```
agatx, line 1220: **P4 — iOS reaches activation/release completion via a bespoke `awaitSettle` poll loop; web/Android await the primitives directly**

iOS made `activateForeground`/`releaseForeground` synchronous, so it infers completion with a two-phase poller (64 yields, then 20ms sleep-against-deadline) plus a separate `awaitForegroundReleaseSettled`. Web and Android make the same primitives async and await them with a single `withTimeout`. The cross-session switch ordering (release-old-before-acquiring-new, contract §6) is thus expressed as direct awaited promises on two platforms and a hand-rolled poll on the third.

The poll thresholds and dual settle signals are iOS-only tuning with no analog elsewhere, so any timing change to the foreground contract must be reasoned about in two unrelated control-flow shapes, and a mismatch between what the poll predicate checks and what `activateForeground` actually sets could let the registry treat a not-yet-settled activation as complete (or hang the serial queue). Consider making the iOS primitives async and awaiting them, to converge the three implementations.

client/packages/core/src/SerenadaCore.ts
```diff
@@ -79,6 +85,45 @@ export class SerenadaCore {
...
+     * `join()` always foregrounds. The (Phase 3) `SerenadaCallRegistry` calls this
+     * to create a `'held'` call (no capture, no coordinator, no lease — stable
+     * senders are still created during negotiation) or a `'foreground'` call.
+     *
+     * For `'foreground'` the caller is responsible for the arbiter lease (the
+     * registry acquires/releases it). For `'held'` no lease is taken. This method
+     * never acquires the lease itself; the public `join()` is the only path that
+     * self-acquires the `direct`-mode lease.
+     */
+    joinInternal(
```
agatx, line 242: **P4 — `join()` and `joinInternal()` are near-duplicate session-construction**

Both contain the same URL-vs-roomId branch (`parseRoomIdFromUrl` or build `roomUrl` from `serverHost`, then `new SerenadaSession(...)`), differing only in per-branch options (`acquireForegroundLease:true` vs `initialMediaRole`). The duplicated branching must be kept in lockstep — a future change to how a room URL maps to a session (new constructor arg, different `roomUrl` fallback, extra option) has to land in both, and an edit applied to only one silently diverges the public single-call path from the registry path. Public `join()` could delegate to `joinInternal({ initialMediaRole: 'foreground' }, ...)` plus the lease flag, collapsing the two copies into one.

client/packages/react-ui/src/hooks/useSerenadaCallRegistry.ts
agatx: Fixed in `a016824b`: removed `config.logger` from the registry `useMemo` deps, so an inline `logger` no longer reconstructs the registry and tears down every live call. The logger is still captured by closure, same accepted tradeoff `useSerenadaSession` makes (it doesn't key on `logger` either). `activeCall` is now memoized too.

client/packages/core/src/SerenadaSession.ts
agatx: Fixed in `a016824b`: `resumeLocalMediaFromHold` now reconciles `this.facingMode` to `desiredVideoMode` (selfie→user, world→environment) before `reacquireVideoTrack`, so a `setCameraMode`/`flipCamera` made while held is applied on resume instead of reacquiring on the stale facing. Parity with iOS/Android.

client/packages/core/src/SerenadaSession.ts
agatx: Fixed in `a016824b`: on the `supersededByGeneration || supersededByToken` branch, `runResume` now throws after rollback when `throwOnFailure` is true (the `activateForeground` caller), so `runSwitch` rolls back instead of treating the superseded resume as a live foreground. `resumeForeground` (`throwOnFailure:false`) is unchanged. Updated the two activateForeground-fencing tests to assert the reject + still-held post-conditions.

client/packages/core/src/media/MediaEngine.ts
agatx: Fixed in `a016824b`: the post-await guard now also re-checks `this.heldNoCapture` and drops/stops the freshly captured track if a hold re-latched during `getUserMedia`, so a held call can no longer briefly publish live mic to peers.

client/packages/core/src/media/MediaEngine.ts
agatx: Fixed in `a016824b`: a re-entrant `reacquireLocalAudioCapture` now awaits the in-flight acquire (body split into a worker + stored promise) and returns after the track lands, so an unmute racing a resume's mic reacquire broadcasts `audioEnabled:true` once the track is attached — no stale `audioEnabled:false`. The `requestingMedia` (initial getUserMedia) case keeps its early-return.

client-android/serenada-core/src/main/java/app/serenada/core/SerenadaCallRegistry.kt
agatx: Fixed in `a016824b`: converged to the web/iOS contract — held calls now publish `mediaActivationState = INACTIVE`, and permission/failure ride solely on `activationError`. Removed the Android-only re-derivation in the snapshot, the three `setMediaActivationState` set-sites, and the now-dead `setMediaActivationState` itself; updated 3 tests to assert the converged contract.

client/packages/react-ui/src/hooks/useSerenadaCallRegistry.ts
agatx: Fixed in `a016824b`: `activeCall` is memoized on the active call's snapshot identity, so its `{state, session}` wrapper identity is stable across unrelated re-renders.

client-android/serenada-core/src/main/java/app/serenada/core/CallRegistryTypes.kt
agatx: Fixed in `a016824b`: extracted a single `extractRoomToken` helper; both `canonicalRoomId` and `SerenadaCore.resolveRoomUrl` now call it, so there's one source of truth for room-token extraction. Each caller's distinct fallback is preserved, so outputs are identical for all inputs.

client-ios/SerenadaCore/Sources/Call/SerenadaAudioCoordinator.swift
agatx: Fixed in `a016824b`: confirmed there are no call sites (repo-wide) and removed `clearForegroundLease(_:)` from the `LeaseAwareAudioCoordinator` protocol and its `DefaultAudioCoordinator` implementation. SerenadaCore still compiles.

client/packages/core/src/SerenadaCallRegistry.ts
agatx: Fixed in `a016824b`: added a comment at both sites (`runLeaveOrEnd` and `handleCallTerminated`) noting the trailing `releaseModeIfIdle()` is load-bearing only on the `endedCallRetentionMs > 0` path (the default path releases via `scheduleEndedRemoval`→`removeCall`).

client/packages/core/src/SerenadaCore.ts
agatx: Fixed in `a016824b`: `join()` now delegates to `joinInternal(room, { initialMediaRole:'foreground', acquireForegroundLease:true, ... })`, so the URL/roomId→session construction lives in one place. `joinInternal` gained an optional `acquireForegroundLease` (the registry leaves it unset). Public `join()` behavior is preserved exactly.

client-ios/SerenadaCore/Sources/SerenadaCallRegistry.swift
agatx: Acknowledged — valid. Deferring: this is part of a cross-platform **publish-contract convergence** (forward held-call fields, coalesced rather than per-tick) that pairs with the web O(N²)-publish and Android `activeCallId`-scan findings. Rather than change publish semantics piecemeal across three SDKs in this PR, we'll land one coalesced contract for all three as a focused follow-up. Current behavior is correct (a held chip can be momentarily stale on iOS until the next op). Leaving open to track.

client/packages/core/src/SerenadaCallRegistry.ts
agatx: Acknowledged — valid (perf only; behavior is correct). Deferring with the publish-contract convergence cluster (iOS under-publish + Android `activeCallId` scan): a single coalesced publish contract — publish the changed call only / skip stats-only emissions of non-active calls — is a deliberate follow-up across all three SDKs, not a piecemeal change in this PR. Leaving open to track.

client-android/serenada-core/src/main/java/app/serenada/core/SerenadaCallRegistry.kt
agatx: Acknowledged — valid (P4, perf only; negligible for small N today). Deferring with the publish-contract convergence cluster: a stored `activeCallId` updated at lease acquire/release transitions (as iOS does) belongs with the publish-coalescing follow-up. Leaving open to track.

client-ios/SerenadaCore/Sources/SerenadaCallRegistry.swift
agatx: Acknowledged — valid. Deferring: making the iOS `activateForeground`/`releaseForeground` primitives async (to converge with the web/Android `withTimeout` await) is a sizable refactor of working, tested code, and the current poll-based settle is correct. Better as a focused follow-up than a risky change in this PR. Leaving open to track.

client-ios/SerenadaCore/Sources/SerenadaCallRegistry.swift
agatx: Addressed in `300c0990`: the iOS registry now observes each session's full state via a new `registrySnapshotPublisher` on the `RegistryManagedSession` seam (objectWillChange-backed) and republishes the changed call, deduped against a cached snapshot — so a held call's participantCount / membershipPhase (waiting→inCall) / actualAudioPublished reflect live, not only at op boundaries. Part of the publish-contract convergence (web/android/iOS now all publish iff a published `ManagedCallState` field changed).

client/packages/core/src/SerenadaCallRegistry.ts
agatx: Addressed in `300c0990`: each managed call caches its last published `ManagedCallState`; a session emission whose recomputed snapshot is unchanged (the ~1s stats tick) now skips `publish()`, eliminating the O(N²) remap + needless re-render. Converged with iOS (now publishes held-call changes) and Android (data-class dedupe + stored activeCallId).

client-android/serenada-core/src/main/java/app/serenada/core/SerenadaCallRegistry.kt
agatx: Addressed in `300c0990`: `activeCallId` is now a stored field updated only via `setForeground`/`clearForeground` (the sole `foregroundToken` writers), so the lookup is O(1) and can't drift from the lease. Also added the per-call snapshot dedupe so `publish()` no longer recomputes every call on each stats tick.

client-ios/SerenadaCore/Sources/SerenadaCallRegistry.swift
agatx: Addressed in `ca21e47c`: iOS `activateForeground`/`releaseForeground` are now async and the registry awaits them with a bounded `withTimeout` — the `awaitSettle` poll is retired for all four foreground call sites (kept only for `awaitHeldJoin`, which has no async primitive to await). This converges iOS with the web/android `withTimeout`-awaited shape. Behavior-preserving: §6 release-before-acquire ordering, both fences (op generation + arbiter owner token), and the release/activate timeout outcomes are unchanged. Verified 87/87 across ForegroundSessionContract / HoldResume / SerenadaCallRegistry / ForegroundMediaArbiter / AudioCoordinatorLeaseFencing.

client-ios/SerenadaCore/Sources/SerenadaSession.swift
```diff
@@ -1134,6 +1484,13 @@ public final class SerenadaSession: ObservableObject {
         let previous = audioCoordinatorLifecycleTask
         let coordinator = audioCoordinator
         let intent = config.audioIntent
+        // The foreground lease this activation runs under (Phase 4, contract §6) so
+        // the default coordinator's own delayed callbacks / observers can fence
+        // against the CURRENT owner. The owner-token id is the registry-issued fence
+        // token (`nil` on the single-call/direct path — one owner, nothing to
+        // fence); the generation is the live op generation.
+        let lease = AudioSessionLease(ownerTokenId: foregroundOwnerToken?.leaseId, generation: mediaOpGeneration)
```
agatx: **P1 · confirmed correctness** — iOS direct calls install a colliding `(nil, 0)` audio lease.

The direct path builds the lease from `foregroundOwnerToken?.leaseId`, which is **always nil** here, so every direct coordinator installs the identical value `(ownerTokenId: nil, generation: 0)` into the process-global `AudioSessionLeaseRegistry`. Android avoids this by binding the unique `directLeaseToken`.

**Failure:** end one 1:1 call and immediately start another. Both install `(nil,0)`. The first call's chained async `deactivateCallSession(fencedBy:)` sees `isLeaseCurrent((nil,0)) == true` (it now matches the *second*, live call) and runs `clearInstalledLease()` → clears the live call's registry lease. The new call's route-change / proximity / interruption observers then see `isLeaseCurrent == false` and silently drop — earpiece/speaker routing and proximity screen-off stop working until something re-activates the session.

Same root cause at `SerenadaCore/Sources/Call/SerenadaAudioCoordinator.swift:219` and `SerenadaSession.swift:1512`. Fix: give the direct path a unique per-session lease token (mirror Android's `directLeaseToken`) instead of leaning on the always-nil `foregroundOwnerToken`.

client-ios/SerenadaCore/Sources/SerenadaSession.swift
```diff
@@ -523,6 +658,21 @@ public final class SerenadaSession: ObservableObject {
...
+        // and wedge future direct joins (I3). `deinit` is nonisolated and the
+        // arbiter is @MainActor, so hop to the main actor — capturing ONLY value
+        // types (the lease token + this session's pre-computed ObjectIdentifier +
+        // the arbiter reference), NEVER `self`, so a deallocating object is not
+        // resurrected into an escaping closure.
+        if let token = directLeaseToken {
+            let arbiter = foregroundArbiter
+            let ownerIdentity = ObjectIdentifier(self)
+            directLeaseToken = nil
+            Task { @MainActor in
```
agatx, line 707: **P1 · confirmed correctness** — cross-room direct join races the deferred deinit lease release.

The direct foreground lease is released from `deinit` via an async `Task { @MainActor }` hop, but `join()` acquires it synchronously-ish inside the async `prepareMediaAndConnect()`. On the single-call host, `CallManager.activateSession(B)` replaces `activeSession` **without** calling `leave()` on A, so A's lease only releases when its deinit Task eventually runs.

**Failure:** user in call A follows a deep link / join-room to call B without first dismissing A. B's `prepareMediaAndConnect()` calls `arbiter.acquireForeground(mode: .direct)` before A's deferred release Task runs → `ForegroundLeaseUnavailable(.leaseLive)` → B fails terminally with "Foreground media unavailable: another call owns it." This is a **new regression** — pre-PR there was no lease and the join succeeded.

Fix: release the outgoing session's lease deterministically before the new direct acquire (e.g. `leave()`/explicit release in `activateSession`), rather than relying on deinit timing.

client/packages/core/src/index.ts
```diff
@@ -83,6 +83,42 @@ export type {
...
+export type {
+    CallMediaRole,
+    MediaActivationState,
+    VideoMode,
+    ForegroundArbiterMode,
+    ForegroundOwnerToken,
+} from './types.js';
+export { ForegroundLeaseUnavailable } from './types.js';
+/** @internal Process-wide foreground media arbiter (one per JS context) */
+export { foregroundArbiter, __resetForegroundArbiterForTests } from './foregroundArbiter.js';
```
agatx: **P2 · confirmed** — the process-singleton arbiter is a real public export.

`foregroundArbiter` (mutable, holds the live single lease/mode) and `__resetForegroundArbiterForTests` are exported as named exports of `@agatx/serenada-core`. The `/** @internal */` tag is cosmetic: no tsconfig enables `stripInternal` (`tsconfig.build.json` doesn't), so both symbols land in the published `.d.ts` **and** runtime JS as fully importable API.

**Failure:** a host (the new samples encourage direct SDK integration) imports `__resetForegroundArbiterForTests` / `foregroundArbiter` and resets or mutates it under a live `SerenadaSession`/`SerenadaCallRegistry`. A later direct join or registry switch then acquires a second foreground lease (two owners of mic/camera/audio-route) or a token-fenced callback no-ops — audio corruption or a wedged `ForegroundLeaseUnavailable` with no recovery.

Fix: don't export these from the package entry. Keep the arbiter internal (e.g. a non-exported module singleton) and expose the test reset only via a deep test-only import path, or enable `stripInternal` and verify the `.d.ts`/JS no longer surface them.

client-android/serenada-core/src/main/java/app/serenada/core/call/SignalingPayloads.kt
```diff
@@ -153,6 +155,7 @@ internal fun JSONObject?.toMediaStatePayload(): MediaStatePayload? {
         fromCid = fromCid,
         audioEnabled = if (has("audioEnabled")) optBoolean("audioEnabled") else null,
         videoEnabled = if (has("videoEnabled")) optBoolean("videoEnabled") else null,
+        held = if (has("held")) optBoolean("held") else null,
```
agatx: **P2 · confirmed parity** — Android coerces a non-boolean `held` to `false`.

`if (has("held")) optBoolean("held")` turns a present-but-non-boolean value (`held: null`, a number, etc.) into `false`, while web (`typeof rec.held === 'boolean' ? ... : undefined`) and iOS (`obj["held"]?.boolValue` → nil) treat it as **unknown**.

**Failure:** if any sender or future server build emits an explicit `null`/non-bool `held` rather than omitting it, a genuinely-held remote participant renders as live/active on Android only — a cross-platform hold-state mismatch. Latent today because the server uses `omitempty`, but the parser shouldn't depend on that.

Fix: only read `held` when it's actually a boolean (mirror web's strict `typeof` check), otherwise leave it `null`.

client/packages/core/src/SerenadaCallRegistry.ts
```diff
@@ -0,0 +1,952 @@
...
+        // call. Everything else is held.
+        const mediaRole: CallMediaRole =
+            call.foregroundToken !== null && call.id === this.activeCallId ? 'foreground' : 'held';
+        return {
+            id: call.id,
+            roomId: call.roomId,
+            roomUrl: call.roomUrl,
+            membershipPhase: callState.phase,
+            mediaRole,
+            mediaActivationState: session.currentMediaActivationState as MediaActivationState,
```
agatx, line 963: **P2 · confirmed parity** — held calls aren't normalized to `INACTIVE` mediaActivationState.

`toManagedCallState` passes `session.currentMediaActivationState` straight through for HELD calls. The Android registry forces `MediaActivationState.INACTIVE` for held calls (`role == FOREGROUND ? ... : INACTIVE`); web does not.

**Failure:** a host with one foreground + one held call reads `registry.calls` and sees the held call reporting a non-`INACTIVE` activation state (whatever the wrapped session last set). A switcher UI that keys "on hold" / activity indicators off `mediaActivationState` renders the held call as still actively capturing — diverging from Android/iOS.

Fix: normalize the held branch to `INACTIVE` to match the cross-platform held contract.

client/packages/core/src/SerenadaSession.ts
```diff
@@ -432,11 +528,65 @@ export class SerenadaSession implements SerenadaSessionHandle {
...
+            this.media.initializeHeldWithoutCapture();
+        } else if (deps.acquireForegroundLease) {
+            // Public single-call `SerenadaCore.join()` routes through the arbiter:
+            // acquire the foreground lease (mode `direct`) BEFORE activating media.
+            // A second concurrent direct join while one is live throws
+            // ForegroundLeaseUnavailable (propagated to the join caller). This is a
+            // SELF-owned lease: record it as `directLeaseToken` so teardown
+            // self-releases it (the registry path never sets this). Also seed the
+            // fence token from the same lease.
+            const token = foregroundArbiter.acquireForeground(this.roomId, 'direct', this);
```
agatx: **P2 · confirmed** — `join()` self-acquires the process-wide direct lease synchronously in the constructor.

When `acquireForegroundLease` is set, the constructor calls `foregroundArbiter.acquireForeground(this.roomId, 'direct', this)` immediately. If a host constructs/joins a new session before the previous one's `destroy()`/`leave()` has released, this throws `ForegroundLeaseUnavailable` synchronously out of `join()`.

**Failure:** a web host that reassigns a session ref without `destroy()`-ing first, or mounts two `SerenadaCallFlow` instances at once, gets an unhandled exception / failed call start instead of the prior "two sessions coexist" behavior. React's `useSerenadaSession`/`SerenadaCallFlow` cleanup (destroy-old before join-new) avoids it; custom hosts and non-effect remount paths don't.

Flagging for intent: if this strictness is desired, document it on the public `join()` contract so integrators know destroy-before-join is mandatory.

client-android/serenada-core/src/main/java/app/serenada/core/CallRegistryTypes.kt
```diff
@@ -0,0 +1,238 @@
...
+/**
+ * Single source of truth for pulling the room token out of a `/call/<token>` URL
+ * (contract §7 "Call identity policy"). Used by BOTH [canonicalRoomId] (registry
+ * dedup key) and [SerenadaCore.resolveRoomUrl] (the join path), so the dedup key
+ * can never disagree with the room the join connects to. Returns the trimmed
+ * last path segment, or null when the input has no usable last segment (callers
+ * apply their own fallback/null handling). Mirrors iOS `DeepLinkParser` and web
+ * `roomIdentity.canonicalizeRoomId`.
+ */
+internal fun extractRoomToken(input: String): String? {
```
agatx, line 65: **P2 · confirmed parity** — three "single source of truth" room-id canonicalizers disagree.

Android `extractRoomToken` uses `Uri.lastPathSegment`; web `canonicalizeRoomId` keys on the `/call/` segment; iOS `DeepLinkParser.extractRoomId` requires `components[0] == "call"`. The docstrings each claim to mirror the others.

**Failure:** for a non-canonical URL like `https://serenada.app/call/<token>/x`, Android returns `x` as the dedup key while web/iOS return `<token>`. The registry then treats the same room as a different call: dedup fails (a second join is allowed where it should be idempotent) and the `held` flag is keyed to the wrong room id. The divergence is invisible at the call site because the docs claim parity.

Fix: align the three on a single rule (require/anchor on the `call` segment) and make the docstrings reflect the actual shared contract.

client/packages/core/src/SerenadaCallRegistry.ts
```diff
@@ -0,0 +1,952 @@
...
+        if (old && old.foregroundToken) {
+            const oldToken = old.foregroundToken;
+            // Mark a release pending so the arbiter refuses any acquire until this
+            // confirms (Core Invariant 1 / contract §2 rule 2).
+            this.arbiter.markReleasePending();
+            const released = await withTimeout(
+                old.session.releaseForeground(oldToken),
+                FOREGROUND_RELEASE_TIMEOUT_MS,
+            );
+            if (!released) {
```
agatx, line 579: **P3 · confirmed (severity flag)** — `runSwitch` timeout path never releases the lease.

`runSwitch` calls `arbiter.markReleasePending()` before draining the old call. On the `FOREGROUND_RELEASE_TIMEOUT_MS` (5s) branch it returns `'failed'` **without** `releaseLease(oldToken)`, so `releasePending` stays `true` and `currentToken` stays `oldToken`.

**Failure:** if draining the active call's media (`releaseForeground` → `suspendLocalMediaForHold`, a `replaceTrack`/stop that never resolves) exceeds 5s, every subsequent registry switch / `joinAndSwitch` **and** every direct `SerenadaCore.join()` (same singleton arbiter) throws `ForegroundLeaseUnavailable(releasePending)` until that call is drained or torn down — the user can't start or switch to any call.

This is cross-platform-consistent and reads as "old keeps the lease," so likely intended — flagging for severity. If intended, consider a recovery path (force-release on terminal teardown) so a stuck drain can't permanently wedge the arbiter.

client/packages/react-ui/src/hooks/useSerenadaCallRegistry.ts
```diff
@@ -0,0 +1,189 @@
...
+            config.turnsOnly,
+            endedCallRetentionMs,
+        ],
+    );
+
+    // Tear down the registry on unmount (or when it is reconstructed): leave
+    // every live call so the foreground lease and owning mode are released.
+    useEffect(() => {
+        return () => {
+            for (const call of registry.state.calls) {
```
agatx: **P3 · plausible** — registry-teardown effect leaves only the last *published* snapshot.

The cleanup iterates `registry.state.calls` (the last published value). A call created/joined but not yet published at unmount — the window after `joinHeld`/`joinAndSwitch` created a managed call but before its first publish lands — is never left.

**Failure:** unmount (or registry reconstruction from a keyed config change) in that window leaves the call's `SerenadaSession` un-left/undestroyed: its signaling WS/SSE stays open as a ghost room participant and its foreground lease/owning-mode can leak, stranding the process in registry mode for later direct joins.

Fix: tear down from the registry's authoritative live set of sessions (including in-flight creations), not the published snapshot.

client/packages/core/src/media/MediaEngine.ts
```diff
@@ -2584,10 +2871,32 @@ export class MediaEngine {
...
+        // guarantee behind held-initial resume (Core Invariant 3 / contract §5):
+        // held senders are reserved `sendrecv` at join, so a resume that lands a
+        // fresh track on them emits NO offer even while the join offer/answer is
+        // still in flight (negotiated `currentDirection` not yet observed). A
+        // transceiver already negotiated `recvonly`/`inactive` (e.g. an answerer
+        // that locally flips to `sendrecv`) still needs a renegotiation, so this
+        // only short-circuits the never-negotiated (`currentDirection == null`)
+        // reserved-send case.
+        if (transceiver.currentDirection === null
+            && (transceiver.direction === 'sendrecv' || transceiver.direction === 'sendonly')) {
```
agatx, line 3751: **P3 · plausible** — `needsLocalTrackNegotiation` short-circuits a never-negotiated reserved-send transceiver.

For a `sendrecv`/`sendonly` transceiver whose `currentDirection === null`, this returns `false` (no renegotiation). That's the intended held-initial-resume optimization, but it assumes the m-line will settle send-capable.

**Failure:** a registry-created held call (`initializeHeldWithoutCapture` reserves audio/video as `sendrecv` with a null track) is resumed while the initial join offer/answer is still in flight. `resumeLocalMediaFromHold` `replaceTrack`s a fresh track onto a transceiver with `currentDirection === null`, so no offer is sent. If negotiation then settles the m-line `recvonly` (peer-driven), the outbound track is attached but never negotiated to send — the call looks live locally while the remote gets nothing until some unrelated renegotiation fires.

Worth a guard: if a resume lands a track during in-flight negotiation, re-check direction after settle and renegotiate if it didn't land send-capable.

client/packages/core/src/SerenadaCore.ts
```diff
@@ -65,20 +66,64 @@ export class SerenadaCore {
...
+            /**
+             * When `true`, the session takes the process-wide foreground lease at
+             * construction. The public `join()` sets this; the registry leaves it
+             * unset and manages the lease itself.
+             */
+            acquireForegroundLease?: boolean;
+        } = { initialMediaRole: 'foreground' },
+    ): SerenadaSession {
+        if (!SerenadaCore.isSupported()) {
+            return this.createUnsupportedSession() as unknown as SerenadaSession;
```
agatx: **P3 · plausible** — the unsupported-browser stand-in handle lacks registry-driven methods.

`joinInternal` returns the unsupported handle cast `as unknown as SerenadaSession`, but that stand-in implements neither the registry methods (`preflightForeground`/`activateForeground`/`releaseForeground`/`abortForegroundActivation`) nor the getters (`currentMediaActivationState`/`currentDesiredAudioEnabled`/`currentMediaRole`/…) the registry calls on every session.

**Failure:** on an unsupported browser, a host using `useSerenadaCallRegistry().joinAndSwitch()`/`switchTo()` drives the registry to call `session.preflightForeground()` on the stand-in → `TypeError` escaping the queued op, and `toManagedCallState` reads `undefined` getters → garbage call-chip state instead of a clean "unsupported" failure.

Fix: have the registry reject unsupported sessions up front, or make the stand-in satisfy the registry contract with safe no-ops/typed failures.

client-android/serenada-core/src/main/java/app/serenada/core/call/DefaultAudioCoordinator.kt
```diff
@@ -156,32 +350,98 @@ internal class DefaultAudioCoordinator(
...
+                "Audio",
+                "Superseded deactivation (a newer same-coordinator activation is live); " +
+                    "full no-op (requestGen=$requestGeneration, currentGen=$leaseGeneration)",
+            )
+            return
+        }
+
+        // Not superseded. Whether the PROCESS-GLOBAL teardown runs turns only on the
+        // OWNER axis (case 2 vs case 3; rationale on [mayTouchGlobalAudioOnDeactivate]).
+        val mayTouchGlobalAudio = mayTouchGlobalAudioOnDeactivate()
```
agatx, line 398: **P3 · plausible** — global audio restore on deactivate is now gated on `hasOtherOwner`.

The previously-unconditional teardown (restore `MODE_NORMAL`, speakerphone/mic, `abandonAudioFocus`) is now behind `mayTouchGlobalAudioOnDeactivate()` = `!ForegroundMediaArbiter.hasOtherOwner(leaseToken)`. The single-call invariant "leaving a call always restores system audio" now depends on that returning false at deactivate time.

**Failure:** if a different direct session has already become the arbiter's current owner before this coordinator's async deactivate runs, `hasOtherOwner` returns true and the whole restore block is skipped — the device is left in `MODE_IN_COMMUNICATION` with focus abandoned only for the stale call, so after the calls end audio stays routed through the earpiece / at call volume.

Worth confirming the ordering guarantee that no foreign owner can appear between drop and async deactivate on the single-call path; otherwise restore state needs to be reconciled when the last owner releases.

client/packages/core/src/SerenadaCallRegistry.ts
```diff
@@ -0,0 +1,952 @@
...
+function generateCallId(): CallId {
+    const c = (globalThis as { crypto?: { randomUUID?: () => string } }).crypto;
+    if (c?.randomUUID) return c.randomUUID();
+    return `call-${Date.now().toString(36)}-${Math.random().toString(36).slice(2, 10)}`;
+}
+
+/**
+ * Race a promise against a timeout. Returns `true` if `promise` fulfills before
+ * the timeout, `false` on timeout. If `promise` REJECTS before the timeout the
+ * rejection propagates out of this function (the race rejects) — callers that
```
agatx, line 1020: **P3 · low-likelihood** — `generateCallId` fallback is non-cryptographic.

When `crypto.randomUUID` is unavailable (older WebView / non-secure context), the fallback is `Date.now()` + `Math.random()`. The registry keys its `calls` Map and `activeCallId` on `CallId`.

**Failure:** two calls created in the same millisecond can in principle collide on `call-<base36ms>-<random>`, so the second join overwrites/aliases the first entry, dropping one call from the switcher. Low real-world likelihood given the 8-char suffix, but it's the only uniqueness guarantee on that path.

Fix: incorporate a monotonic per-registry counter into the fallback id.

client-ios/SerenadaCore/Sources/SerenadaCallRegistry.swift
```diff
@@ -0,0 +1,1169 @@
...
+    /// in `calls` until the next op or terminal phase). This is a STATE REFRESH,
+    /// NOT an operation: it is deliberately NOT routed through `runQueued` (it
+    /// touches no lease/call-map invariant, only re-derives a published value) and
+    /// it dedupes against the call's `lastSnapshot` so a stats tick that does not
+    /// change the projection is a no-op (no `calls` churn).
+    ///
+    /// The emit hops to the NEXT main-actor turn because the concrete session maps
+    /// `objectWillChange`, which fires in `willSet` BEFORE the property updates;
+    /// reading the snapshot synchronously would observe the OLD value.
+    private func observeStateForPublish(_ call: ManagedCall) {
```
agatx, line 743: **P3 · efficiency** — a `Task { @MainActor }` is spawned per `objectWillChange` emission.

`observeStateForPublish` spawns a fresh Task on every `@Published` change — including the ~1s stats tick on every held call — only to dedupe-and-drop in `republishIfChanged`.

**Failure (steady-state cost):** with N managed calls, the per-second stats tick fires `objectWillChange` on each session, each spawning a Task that hops a main-actor turn and recomputes a full snapshot just to discard it via the `lastSnapshot` dedupe. The dedupe avoids the publish but not the per-tick Task allocation + hop + recompute — O(N) Tasks/sec for zero state change.

Same shape on web at `SerenadaCallRegistry.ts:302` and `:666`. Fix: coalesce emissions (or capture only changed fields into a lighter path) so unchanged ticks don't spawn work.

client/packages/core/src/constants.ts
```diff
@@ -18,6 +18,22 @@ export const JOIN_CONNECT_KICKSTART_MS = 1200;
...
+//
+// Bounds draining the old foreground session's resources (suspend/replace
+// senders, mute playout) before a switch releases its lease.
+export const FOREGROUND_RELEASE_TIMEOUT_MS = 5000;
+// Bounds a new foreground activation. Deliberately greater than
+// `AUDIO_COORDINATOR_TIMEOUT_MS` (10000) so the inner audio-coordinator timeout
+// fires first and the activation surfaces that as the cause.
+export const FOREGROUND_ACTIVATE_TIMEOUT_MS = 12000;
+// Bounds a held room join performed outside the operation queue; mirrors
+// `JOIN_HARD_TIMEOUT_MS`.
```
agatx, line 34: **P3 · DRY** — `HELD_JOIN_TIMEOUT_MS` duplicates `JOIN_HARD_TIMEOUT_MS`.

`HELD_JOIN_TIMEOUT_MS = 15000` with a comment saying it "mirrors `JOIN_HARD_TIMEOUT_MS`" (also 15000) — but nothing links the two values. The same duplicated literal is copied to the Android (`heldJoinTimeoutMs`) and iOS (`HELD_JOIN_TIMEOUT_MS`) resilience files, and consumed at `SerenadaCallRegistry.ts:514`/`:819` and `SerenadaSession.ts:1336`.

**Failure:** retuning `JOIN_HARD_TIMEOUT_MS` silently leaves the held-join bound at the old value across all three platforms, so held vs direct joins drift apart despite the doc promising they track each other.

Fix: derive `HELD_JOIN_TIMEOUT_MS` from `JOIN_HARD_TIMEOUT_MS` rather than hard-coding the same literal.

client-ios/SerenadaCore/Sources/SerenadaSession.swift
agatx: Fixed in `87693fde`: the direct-path audio lease now binds the session's own unique `directLeaseToken` id (`ownerTokenId: foregroundOwnerToken?.leaseId ?? directLeaseToken?.leaseId`) instead of a shared `nil`, so two distinct direct sessions install distinct leases — parity with Android binding `directLeaseToken`. iOS lease-fence + registry + hold/resume suites stay green (87/87). This alone is latent (the `isCurrent(nil)` short-circuit + single-live-direct-lease invariant masked it); the cross-room switch window it pairs with is tracked in the SerenadaSession.swift:672 thread.

client/packages/core/src/index.ts
agatx: Fixed in `87693fde`: removed the `foregroundArbiter` singleton + `__resetForegroundArbiterForTests` re-exports from the package entry (`index.ts`) — with no `stripInternal` they were landing in the published `.d.ts` and JS bundle. No app/sample/react-ui code imported them from the barrel; the one consumer (the react-ui hook test) now imports the reset from `core/src/foregroundArbiter.js` directly, like the core tests. Web suite green (669).

client-android/serenada-core/src/main/java/app/serenada/core/call/SignalingPayloads.kt
agatx: Fixed in `87693fde`: `held` is now read strictly (`opt("held") as? Boolean`) on both the media-state and participant-list paths, so a present-but-non-boolean value stays unknown (null) rather than coercing to false — parity with web (`typeof === 'boolean'`) and iOS (`boolValue`). Added parse tests for the non-boolean case. Scoped to `held` (the field this PR adds); `audioEnabled`/`videoEnabled` keep their existing `optBoolean` read. (Note: at the public model both Android and iOS already collapse unknown→false via `?: false`; only web preserves nullable `held`. This aligns the parse layer.)

client-android/serenada-core/src/main/java/app/serenada/core/CallRegistryTypes.kt
agatx: Fixed in `87693fde`: `extractRoomToken` now keys on the segment after `call/` (falling back to the last segment for non-`/call/` URLs) instead of `lastPathSegment`, so a non-canonical `/call/<token>/extra` resolves to `<token>` — matching web `roomIdentity` and iOS `DeepLinkParser`, and keeping the registry dedup key in lockstep with the join path (both call this one function). Added tests including a `canonicalRoomId(url) == canonicalRoomId(id)` agreement case. Real Serenada URLs are `/call/<token>`, so this is hardening for the non-canonical edge the docstring claimed parity on.

client/packages/react-ui/src/hooks/useSerenadaCallRegistry.ts
agatx: Fixed in `87693fde`: added `SerenadaCallRegistry.close()` — it leaves every call from the authoritative `calls` map (not the last published snapshot) and sets a `closed` flag so a queued create that lands after teardown (an in-flight `joinHeld`/`joinAndSwitch` whose body runs post-unmount) no-ops instead of leaking a session nothing would leave. The react-ui hook cleanup now calls `close()`. Added registry tests for close()-leaves-all, the in-flight-create-after-close no-op, and closed-rejects-join.

client/packages/core/src/SerenadaCore.ts
agatx: Fixed in `87693fde`: `joinInternal` now throws on an unsupported browser instead of laundering the unsupported stand-in (`as unknown as SerenadaSession`). The registry's `createOrReuseCall` already wraps `createSession` in try/catch, so this surfaces as a normal `joinFailed` (mode claim released, no broken managed call published) rather than a `TypeError` when the registry calls a missing method. Public `join()` still returns the stand-in handle (it is a `SerenadaSessionHandle`). Added a throwing-factory registry test.

client/packages/core/src/SerenadaCallRegistry.ts
agatx: Fixed in `87693fde`: the non-crypto `generateCallId` fallback now appends a module-level monotonic counter, so two ids minted in the same millisecond can't collide on the `calls` Map key on engines lacking `crypto.randomUUID`.

client/packages/core/src/SerenadaCallRegistry.ts
agatx: Evaluated — not a divergence in practice. Web's `toManagedCallState` passes `session.currentMediaActivationState` through, and so does iOS (`managedSession.mediaActivationState`); only Android forces `INACTIVE`, which its own comment documents as a deliberate add-on over the web/iOS passthrough. A held web session already reports `inactive` (set at held-initial construction, on hold, and on superseded/failed-resume rollback); the transient `activating`/`active` states only occur while the call's role is already `foreground`. So for a held call web == iOS == `inactive` observably. Forcing it on web would trade one documented divergence (vs Android) for another (vs iOS) without changing output. Leaving the passthrough; resolving.

client/packages/core/src/SerenadaCallRegistry.ts
agatx: Evaluated — not a permanent wedge. `releaseLease` clears `currentToken` and `releasePending` together, so `(currentToken=null, releasePending=true)` is unreachable: `releasePending=true` always implies `old` still owns the lease. Every recovery op (`runSwitch`/`runHold`/`runLeaveOrEnd`) re-drains the current owner (still `old`) and calls `releaseLease` BEFORE any `acquireForeground`, which clears the flag — so the next switch/hold/leave self-heals. This matches the iOS design, which documents exactly this (SerenadaCallRegistry.swift:755-759: “the still-set releasePending flag is cleared by the next switch retry”). The flag is redundant-but-consistent on the abort path, not a leak. Resolving.

client/packages/core/src/media/MediaEngine.ts
agatx: Evaluated — not reachable in Serenada's negotiation. The `currentDirection===null && direction∈{sendrecv,sendonly}` short-circuit is the documented held-initial-resume optimization. On resume, `replaceAudioTrackOnAllPeers` re-checks `needsLocalTrackNegotiation` AFTER `replaceTrack` and force-flips `direction='sendrecv'` + `negotiationNeeded` if the m-line isn't send-capable, so a settled-non-send line still renegotiates. The only gap (currentDirection still null AND direction sendrecv at resume, then the join answer settles non-send with nothing re-evaluating) requires a peer answering recvonly to a sendrecv audio offer — which Serenada's symmetric 1:1 peers never do. The comment already documents the invariant. Leaving as-is; resolving.

client-android/serenada-core/src/main/java/app/serenada/core/call/DefaultAudioCoordinator.kt
agatx: Evaluated — safe for single-call. `hasOtherOwner(token)` returns false when there's no current owner OR when the current owner IS this token (`live !== token`, with a null short-circuit). On single-call teardown the SYNC `deactivate()` runs while the direct lease is still held (`currentToken === directLeaseToken`), so `hasOtherOwner` is false and MODE_NORMAL/route/focus ARE restored; the lease is released afterward and the async deactivate body is then an idempotent no-op. Android tokens are reference-identity (minted per acquire, compared with `!==`), so there's no value-collision analog to the iOS `(nil,0)` concern. The gate only suppresses the global restore in a genuine cross-coordinator handoff (a different live owner), which is intended. Resolving.

client/packages/core/src/constants.ts
agatx: Won't change — the DRY refactor would weaken the cross-platform guard. `scripts/check-resilience-constants.mjs` matches `export const NAME = <number>` (numeric literal only); rewriting `HELD_JOIN_TIMEOUT_MS = JOIN_HARD_TIMEOUT_MS` makes the RHS an identifier, drops it from the script's TS map, and silently removes web from the parity check for that constant. Android (`WebRtcResilienceConstants.kt`) and iOS (`...swift`) also duplicate the literal with the same “mirrors JOIN_HARD_TIMEOUT_MS” comment by design — the literal IS what the guard enforces. Resolving.

client-ios/SerenadaCore/Sources/SerenadaCallRegistry.swift
agatx: Evaluated — declining the coalescing change. The cost is small: held calls fire `objectWillChange` ~2×/s via the stats tick, each spawning a Task that does a struct compare in `republishIfChanged` and returns (the publish-iff-changed dedupe already drops it). For typical N that's a handful of trivially-short Tasks/sec. The proposed fix (`receive(on:)` + sync pre-check) changes the timing of the just-converged publish pipeline and risks the publish-contract test suite (stub sessions fire a Combine subject), for a negligible steady-state gain. Keeping the correct-and-simple version; resolving. Can revisit if held-call counts grow.

client-ios/SerenadaCore/Sources/SerenadaSession.swift
agatx: Confirmed and valid. The blast-radius reducer is landed in `87693fde` (the direct lease now has a unique id, so a stale session's release can't clobber a new owner's lease — see the SerenadaSession.swift:1492 thread). The remaining race — a single-call host (`CallManager.activateSession`) swapping `activeSession` A→B WITHOUT `leave()`ing A, so A's deinit-deferred async lease release can land after B's acquire (or B's acquire throws `.leaseLive`) — is a host-integration fix (synchronous teardown-before-switch). The host apps were reverted to single-call references at `main` as part of productization, so re-diverging them needs a deliberate cross-platform (iOS/Android/web) pass + host tests. Leaving open to track as a focused follow-up.

client/packages/core/src/SerenadaSession.ts
agatx: Confirmed. `join()`'s synchronous `acquireForeground` is intentional fast-fail (the `SerenadaCore.join` doc: a second concurrent direct join “fails fast with ForegroundLeaseUnavailable”), but the point stands — it's a synchronous throw out of a method typed to return a handle, so a host switching rooms without `leave()` first must try/catch or it crashes. Surfacing this as a `CallState` error (defer the acquire into `start()`, add a `foregroundUnavailable` code) instead of a constructor throw is the ergonomic fix; it changes the public `join()` contract and pairs with the host teardown-before-switch work in the SerenadaSession.swift:672 thread. Leaving open to track as the same focused follow-up.

client-ios/SerenadaCore/Sources/SerenadaSession.swift
agatx: Fixed in `fab75ee6`. The user-facing manifestation — the single-call bundled apps switching to a DIFFERENT room (e.g. a deep link) without leaving the current call, so the new join failed on the still-held lease — is resolved by host teardown-before-switch: iOS `CallManager.activateSession` and Android `CallManager.joinRoom` now leave/cancel the previous session before the new one acquires the lease. Web was already safe (React effect cleanup destroys the old session before the new join). The deinit release stays only as a backstop for sessions dropped without leave(); combined with the unique direct-lease id (`87693fde`), a stale deinit release can no longer clobber a live owner. iOS suite 89/89. Resolving.

client/packages/core/src/SerenadaSession.ts
agatx: Update: the room-switch regression this pairs with is fixed in `fab75ee6` (host teardown-before-switch), and the web bundled app is verified safe — navigating between rooms unmounts the old `CallRoom`, whose effect cleanup `destroy()`s the session (releasing the lease) before the new `join()`. So neither bundled app hits this. What remains is purely an SDK-ergonomics footgun for THIRD-PARTY hosts: `join()` is typed to return a handle but acquires the lease synchronously, so a consumer who joins a second room without leaving the first gets a synchronous throw rather than a `CallState` error. Surfacing it as a `CallState` error (defer the acquire to `start()`, add a `foregroundUnavailable` code) is a non-blocking fast-follow. Leaving open to track.

agatx: Fixed in `ae42fb7e`. Web direct `join()` no longer throws synchronously on lease-unavailable: the acquire is deferred from the constructor to `start()`, and on failure the session surfaces an error `CallState` (`failWithError`, code `unknown`, message "Foreground media unavailable: another call owns it") instead of throwing — so `join()` always returns a usable handle. This converges web with iOS/Android, which already catch the arbiter throw at the acquire site and call `handleError`. The arbiter primitive `acquireForeground` still throws; only the public `join()` path no longer propagates it. Updated the two tests that asserted the throw + docs (integration guide + contract). Web suite 669/669. Resolving.

PR #138, opened 2026-06-29
IOS-4409: Add onMediaPermissionBlocked callback for denied in-call media toggles

danyellz:
When a user toggles camera or microphone during a call but the system permission is denied or restricted, the session previously failed silently. Add an `onMediaPermissionBlocked` callback that fires in those cases so host apps can surface a prompt steering the user to Settings, matching the guidance shown on Android.

- Camera: fired from `ensureCameraPermissionForVideoEnable` in the `.denied`/`.restricted` case.
- Microphone: `toggleAudio()` now guards on the audio authorization status and fires for `.denied`/`.restricted`. `.notDetermined` is unaffected — the session still prompts for access.

Client Support PR https://github.com/zelloptt/zello-ios-client/pull/10987

### DIFF
--- a/client-ios/SerenadaCore/Sources/SerenadaSession.swift
+++ b/client-ios/SerenadaCore/Sources/SerenadaSession.swift
@@ -175,14 +175,12 @@ public final class SerenadaSession: ObservableObject {
         return true
     }
 
-    /// Callback invoked when camera/microphone permissions are needed before joining.
+    /// Callback invoked when camera/microphone permissions are needed — either before joining, or
+    /// when a user-initiated in-call toggle (enabling video, unmuting) requires a permission that
+    /// isn't currently usable. Mirrors the Android SDK, which routes both cases through this same
+    /// callback. Hosts disambiguate join vs. in-call via `state.phase` (`.awaitingPermissions`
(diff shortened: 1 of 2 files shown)


client-ios/SerenadaCore/Sources/SerenadaSession.swift
```diff
@@ -178,6 +178,12 @@ public final class SerenadaSession: ObservableObject {
     /// Callback invoked when camera/microphone permissions are needed before joining.
     public var onPermissionsRequired: (([MediaCapability]) -> Void)?
 
+    /// Callback invoked when a user-initiated in-call media toggle (camera or microphone) cannot
+    /// proceed because the corresponding system permission is denied or restricted. The host
+    /// should surface a prompt steering the user to Settings. It is NOT fired for the
+    /// `.notDetermined` state — the session prompts for access in that case.
+    public var onMediaPermissionBlocked: ((MediaCapability) -> Void)?
```
agatx: Instead of a new callback we should probably just fire `onPermissionsRequired`. That would be consistent with how Android works today.

agatx (approved):

PR #142, opened 2026-07-13
ANDROID-3860 Move terminal media teardown off main thread

agatx:
## Summary

- detach session-visible peer and local-media state synchronously during `leave()`
- move terminal camera/media disposal, peer connection close/dispose, and factory/audio-device release to the existing `serenada-pc-dispose` worker
- move default SDK audio-session deactivation off Main while preserving Main-thread behavior for host-provided audio coordinators
- add coverage for deferred terminal peer teardown and dispose-queue ordering

## Root cause

Terminal `SerenadaSession.leave()` synchronously performed native WebRTC and audio teardown on Main. Camera stop/dispose, peer connection close/dispose, and factory/audio-device release can wait on native media threads for hundreds of milliseconds, starving UI frames and exposing a black call-exit frame in Zello Android.

## Impact

`leave()` still transitions the session to `CallPhase.Idle` synchronously, but terminal native resource release may complete asynchronously. This keeps renderer/session state deterministic without blocking the host application's main thread.

## Validation

- `./gradlew :serenada-core:testDebugUnitTest`
- `./gradlew :serenada-core:lintDebug`
- physical-device video teardown on Galaxy S8 (Android 9) and A065 (Android 16): no teardown-time skipped-frame warnings
- locally published SDK integrated into Zello Android 7.13 debug on both devices
- Zello gated and ungated end-to-end video-call teardown recordings: no black exit frame and no teardown-time skipped-frame warning

A separate S8 activation-time frame skip was observed before call connection; it is outside the terminal teardown path changed here.

Jira: [ANDROID-3860](https://zelloptt.atlassian.net/browse/ANDROID-3860)

### DIFF
--- a/client-android/serenada-core/src/main/java/app/serenada/core/SerenadaSession.kt
+++ b/client-android/serenada-core/src/main/java/app/serenada/core/SerenadaSession.kt
@@ -60,6 +60,9 @@ import app.serenada.core.call.SignalingMessage
 import app.serenada.core.call.WebRtcEngine
 import app.serenada.core.call.WebRtcResilienceConstants
 import app.serenada.core.call.CameraCaptureController
+import app.serenada.core.call.PROCESS_TEARDOWN_HANDOFF_TIMEOUT_MS
+import app.serenada.core.call.defaultAudioTeardownFence
+import app.serenada.core.call.terminalMediaTeardownFence
 import app.serenada.core.call.toContentStatePayload
 import app.serenada.core.network.CoreApiClient
 import app.serenada.core.network.SessionAPIClient
@@ -73,7 +76,6 @@ import kotlinx.coroutines.cancelChildren
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.withContext
 import kotlinx.coroutines.withTimeout
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
@@ -1311,6 +1313,22 @@ class SerenadaSession internal constructor(
 
         acquirePerformanceLocks()
         providerScope.launch {
+            val previousMediaReleased =
+                terminalMediaTeardownFence.awaitPending(PROCESS_TEARDOWN_HANDOFF_TIMEOUT_MS)
+            val previousAudioRestored =
+                defaultAudioTeardownFence.awaitPending(PROCESS_TEARDOWN_HANDOFF_TIMEOUT_MS)
+            if (!previousMediaReleased || !previousAudioRestored) {
+                logger?.log(
+                    SerenadaLogLevel.ERROR,
+                    "Session",
+                    "Previous call teardown timed out before media activation",
+                )
+                handleError(CallError.Unknown("Previous call teardown timed out"))
+                return@launch
+            }
+            if (_state.value.phase != CallPhase.Joining || joinFlowCoordinator.joinAttemptSerial != joinAttemptId) {
+                return@launch
+            }
             try {
                 withTimeout(WebRtcResilienceConstants.AUDIO_COORDINATOR_TIMEOUT_MS) {
                     audioCoordinatorMutex.withLock {
@@ -1327,7 +1345,14 @@ class SerenadaSession internal constructor(
                 handleError(CallError.Unknown(e.message ?: "Audio session activation failed"))
                 return@launch
             }
-            if (!isActive) return@launch
+            if (
+                !isActive ||
+                _state.value.phase != CallPhase.Joining ||
+                joinFlowCoordinator.joinAttemptSerial != joinAttemptId
+            ) {
+                runCatching { audioCoordinator.deactivateCallSession() }
+                return@launch
+            }
             joinFlowCoordinator.scheduleJoinTimeout(roomId, joinAttemptId)
             try {
                 callAudioSessionController.activate()
@@ -2241,9 +2266,9 @@ class SerenadaSession internal constructor(
         joinFlowCoordinator.reset()
         peerNegotiationEngine.resetAll()
         iceFetchGeneration += 1
-        // Host-provided session controllers keep their established Main-thread callback. The
-        // SDK default coordinator is deactivated below on a worker because AudioManager route
-        // restoration can synchronously block Main for hundreds of milliseconds on some devices.
+        // Host-provided session controllers keep their established Main-thread callback. The SDK
+        // default coordinator synchronously detaches Main-owned callbacks inside
+        // deactivateCallSession(), then performs only the blocking AudioManager restore off Main.
         if (callAudioSessionController !== defaultAudioCoordinator) {
             callAudioSessionController.deactivate()
         }
@@ -2251,13 +2276,7 @@ class SerenadaSession internal constructor(
             try {
                 withTimeout(WebRtcResilienceConstants.AUDIO_COORDINATOR_TIMEOUT_MS) {
                     audioCoordinatorMutex.withLock {
-                        if (audioCoordinator === defaultAudioCoordinator) {
-                            withContext(Dispatchers.Default) {
-                                audioCoordinator.deactivateCallSession()
-                            }
-                        } else {
-                            audioCoordinator.deactivateCallSession()
-                        }
+                        audioCoordinator.deactivateCallSession()
                     }
                 }
             } catch (e: TimeoutCancellationException) {
--- a/client-android/serenada-core/src/main/java/app/serenada/core/SerenadaSession.kt
+++ b/client-android/serenada-core/src/main/java/app/serenada/core/SerenadaSession.kt
@@ -61,7 +61,7 @@ import app.serenada.core.call.WebRtcEngine
 import app.serenada.core.call.WebRtcResilienceConstants
 import app.serenada.core.call.CameraCaptureController
 import app.serenada.core.call.PROCESS_TEARDOWN_HANDOFF_TIMEOUT_MS
-import app.serenada.core.call.defaultAudioTeardownFence
+import app.serenada.core.call.audioCoordinatorTeardownFence
 import app.serenada.core.call.terminalMediaTeardownFence
 import app.serenada.core.call.toContentStatePayload
 import app.serenada.core.network.CoreApiClient
@@ -1316,7 +1316,7 @@ class SerenadaSession internal constructor(
             val previousMediaReleased =
                 terminalMediaTeardownFence.awaitPending(PROCESS_TEARDOWN_HANDOFF_TIMEOUT_MS)
             val previousAudioRestored =
-                defaultAudioTeardownFence.awaitPending(PROCESS_TEARDOWN_HANDOFF_TIMEOUT_MS)
+                audioCoordinatorTeardownFence.awaitPending(PROCESS_TEARDOWN_HANDOFF_TIMEOUT_MS)
             if (!previousMediaReleased || !previousAudioRestored) {
                 logger?.log(
                     SerenadaLogLevel.ERROR,
@@ -1350,7 +1350,9 @@ class SerenadaSession internal constructor(
                 _state.value.phase != CallPhase.Joining ||
                 joinFlowCoordinator.joinAttemptSerial != joinAttemptId
             ) {
-                runCatching { audioCoordinator.deactivateCallSession() }
+                val deactivationJob =
+                    audioCoordinatorDeactivationJob ?: beginAudioCoordinatorDeactivation()
+                deactivationJob.join()
                 return@launch
             }
             joinFlowCoordinator.scheduleJoinTimeout(roomId, joinAttemptId)
@@ -2272,21 +2274,7 @@ class SerenadaSession internal constructor(
         if (callAudioSessionController !== defaultAudioCoordinator) {
             callAudioSessionController.deactivate()
         }
-        val deactivationJob = audioCoordinatorScope.launch {
-            try {
-                withTimeout(WebRtcResilienceConstants.AUDIO_COORDINATOR_TIMEOUT_MS) {
-                    audioCoordinatorMutex.withLock {
-                        audioCoordinator.deactivateCallSession()
-                    }
-                }
-            } catch (e: TimeoutCancellationException) {
-                logger?.log(SerenadaLogLevel.WARNING, "Audio", "Audio session deactivation timed out")
-            } catch (e: Exception) {
-                if (e is kotlinx.coroutines.CancellationException) throw e
-                logger?.log(SerenadaLogLevel.WARNING, "Audio", "Failed to deactivate audio session: ${e.message}")
-            }
-        }
-        audioCoordinatorDeactivationJob = deactivationJob
+        val deactivationJob = beginAudioCoordinatorDeactivation()
         releasePerformanceLocks()
         stopRemoteVideoStatePolling()
         signalingProvider.disconnect()
@@ -2323,6 +2311,35 @@ class SerenadaSession internal constructor(
         if (clearRecovery) recoveryStorage.clear()
         providerScope.coroutineContext.cancelChildren()
         updateDiagnostics(CallDiagnostics())
+        return deactivationJob
+    }
+
+    private fun beginAudioCoordinatorDeactivation(): Job {
+        val teardownTicket = audioCoordinatorTeardownFence.begin()
+        val deactivationJob = audioCoordinatorScope.launch {
+            try {
+                if (!teardownTicket.awaitTurn(PROCESS_TEARDOWN_HANDOFF_TIMEOUT_MS)) {
+                    logger?.log(
+                        SerenadaLogLevel.WARNING,
+                        "Audio",
+                        "Timed out waiting for the previous audio coordinator teardown",
+                    )
+                }
+                withTimeout(WebRtcResilienceConstants.AUDIO_COORDINATOR_TIMEOUT_MS) {
+                    audioCoordinatorMutex.withLock {
+                        audioCoordinator.deactivateCallSession()
+                    }
+                }
+            } catch (e: TimeoutCancellationException) {
+                logger?.log(SerenadaLogLevel.WARNING, "Audio", "Audio session deactivation timed out")
+            } catch (e: Exception) {
+                if (e is kotlinx.coroutines.CancellationException) throw e
+                logger?.log(SerenadaLogLevel.WARNING, "Audio", "Failed to deactivate audio session: ${e.message}")
+            } finally {
+                teardownTicket.complete()
+            }
+        }
+        audioCoordinatorDeactivationJob = deactivationJob
         return deactivationJob
     }
 
--- a/client-android/serenada-core/src/main/java/app/serenada/core/SerenadaSession.kt
+++ b/client-android/serenada-core/src/main/java/app/serenada/core/SerenadaSession.kt
@@ -73,6 +73,7 @@ import kotlinx.coroutines.cancelChildren
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import kotlinx.coroutines.withTimeout
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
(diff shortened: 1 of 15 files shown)


client-android/serenada-core/src/main/java/app/serenada/core/call/WebRtcEngine.kt
```diff
@@ -350,37 +359,66 @@ internal class WebRtcEngine(
         videoSource = null
     }
 
-    fun stopLocalMedia() {
+    private fun detachLocalMediaForRelease(): LocalMediaResources {
         cameraController.resetCameraState()
-        screenShareController.reset()
         localVideoTrack?.setEnabled(false)
         localAudioTrack?.setEnabled(false)
+        localVideoTrack?.let { track ->
```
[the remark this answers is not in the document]
agatx: Fixed in `23cae965`: `detachLocalMediaForRelease()` now disables `localContentVideoTrack` synchronously alongside camera and audio before queuing native teardown. The full `:serenada-core:testDebugUnitTest`, `:serenada-core:lintDebug`, and `:app:assembleDebug` checks pass.

client-android/serenada-core/src/main/java/app/serenada/core/call/WebRtcEngine.kt
```diff
@@ -350,37 +359,67 @@ internal class WebRtcEngine(
...
-        stopLocalMedia()
-        localSinks.clear()
-        localContentSinks.clear()
-        peerConnectionDisposeQueue.flush(shutdownAfterDrain = true) {
+        val localMedia = detachLocalMediaForRelease()
+        // Capturer/track/source disposal, PeerConnection.close/dispose, and the final
+        // factory/ADM release synchronously wait on media and libwebrtc threads on some devices.
+        // Sinks and session-visible state have already been detached on Main, so the remaining
+        // native teardown can run without racing renderer unmounts or stale session callbacks.
+        peerConnectionDisposeQueue.post {
```
agatx: **[P1] (confidence: 8/10) Fast leave-then-rejoin can silently downgrade the next call to audio-only.**

`release()` returns immediately, but the old capturer's `stopCapture()`/`dispose()` runs on the dispose thread *after* all the peer connection closes drain, and those closes are the hundreds-of-ms stalls this PR is about. `startJoinInternal()` builds a new engine and `startLocalMedia()` opens the camera with no fence against that in-flight teardown. A quick rejoin inside the window hits `CAMERA_IN_USE` (worse in COMPOSITE mode, which holds two cameras), `restartVideoCapturerWithFallback` exhausts its candidates, and the join continues with only `"No camera capturer available; continuing audio-only"` in the log. No retry, nothing surfaced to the host. The same window applies to the mic: the old ADM is released in this runnable while the new engine has already created its own.

Before this PR the capturer was stopped synchronously inside `leave()`, so the camera was always free before a rejoin could start.

Suggested fix: expose a teardown-complete signal from `release()` (the queue already supports `flush(onDrained = ...)`), stash it where the next session can see it (e.g. on `SerenadaCore` or a companion), and have the next `startLocalMedia` await it off Main with a bounded timeout. Stopping just the capturer synchronously in `detachLocalMediaForRelease()` also closes the camera half, but trades back some of the jank this PR removes, so the fence is the better shape.

(Found independently by 3 of 4 review passes.)

client-android/serenada-core/src/main/java/app/serenada/core/SerenadaSession.kt
```diff
@@ -2235,12 +2241,23 @@ class SerenadaSession internal constructor(
...
+        if (callAudioSessionController !== defaultAudioCoordinator) {
+            callAudioSessionController.deactivate()
+        }
         val deactivationJob = audioCoordinatorScope.launch {
             try {
                 withTimeout(WebRtcResilienceConstants.AUDIO_COORDINATOR_TIMEOUT_MS) {
                     audioCoordinatorMutex.withLock {
-                        audioCoordinator.deactivateCallSession()
+                        if (audioCoordinator === defaultAudioCoordinator) {
+                            withContext(Dispatchers.Default) {
```
agatx: **[P1] (confidence: 7/10) Cross-session race: the old coordinator's async restore can clobber the next call's audio.**

`audioCoordinatorMutex` is per session, and `SerenadaCore.join()` constructs a brand-new session with its own `DefaultAudioCoordinator`, so nothing serializes this worker-thread `deactivate()` against the next session's `activate()`. Before this PR that was impossible: `audioCoordinatorScope` is `Main.immediate` and the mutex is uncontended at this point, so the deactivation completed synchronously inside `leave()` before it returned.

Two bad orderings on a fast leave-then-join, both against process-global `AudioManager` state:
1. The new session's `activate()` runs first and snapshots `previousAudioMode` while the old call's `MODE_IN_COMMUNICATION` is still set, so in-call mode becomes the "previous" state and never restores after the new call ends.
2. The old `deactivate()` finishes after the new `activate()` and knocks the live call back to `MODE_NORMAL`, clears the communication device, and restores stale speakerphone/mute (DefaultAudioCoordinator.kt:159-186).

Suggested fix: a process-wide guard for the default coordinator, e.g. a companion-level activation generation. `activate()` bumps it and records it, `deactivate()` skips the AudioManager restore (but still abandons focus and unregisters listeners) if the generation moved on. That avoids cross-session locking entirely.

Related, smaller: the mutex is held across this blocking `withContext(Dispatchers.Default)` block and `withTimeout` cannot interrupt it, so a pathological AudioManager hang (>10s) now fails the *next* join with "Audio session activation timed out" where pre-PR it just janked. Acceptable, but worth knowing the failure moved onto the join path.

client-android/serenada-core/src/main/java/app/serenada/core/SerenadaSession.kt
```diff
@@ -2235,12 +2241,23 @@ class SerenadaSession internal constructor(
         joinFlowCoordinator.reset()
         peerNegotiationEngine.resetAll()
         iceFetchGeneration += 1
-        callAudioSessionController.deactivate()
+        // Host-provided session controllers keep their established Main-thread callback. The
+        // SDK default coordinator is deactivated below on a worker because AudioManager route
+        // restoration can synchronously block Main for hundreds of milliseconds on some devices.
+        if (callAudioSessionController !== defaultAudioCoordinator) {
```
agatx, line 2274: **[P2] (confidence: 7/10) `DefaultAudioCoordinator` becomes multi-threaded here, but its state is unsynchronized Main-owned vars.**

Until this PR the class was Main-confined by construction: the proximity listener, `AudioDeviceCallback`, focus-change `handler.post`, and the delayed route refresh are all registered on the Main `handler`, and every guard is a plain read of `audioSessionActive` / `proximityMonitoringActive` / `isProximityNear` (DefaultAudioCoordinator.kt:38-46, no `@Volatile`, no lock). `deactivate()` now writes all of those from a `Dispatchers.Default` worker.

The realistic trigger is hang-up itself: the user pulls the phone off their ear to tap End, the proximity transition is delivered on Main right as `deactivate()` runs on the worker, and `onSensorChanged` -> `applyCallAudioRouting()` (DefaultAudioCoordinator.kt:113-125) re-applies a call route mid-restore because its `audioSessionActive` read is stale or interleaved. Queued device-change and focus callbacks have the same window. Result: the device can exit the call with communication mode or speakerphone left in the wrong state.

Suggested fix, mirroring what this PR already did for the engine: split deactivation into a synchronous Main-side detach (flip `audioSessionActive`, `stopProximityMonitoring()`, `stopAudioDeviceMonitoring()`, clear pinned route, snapshot the previous-mode values) and keep only the blocking AudioManager restore on the worker. If that feels heavy, `@Volatile` on the guard flags is the cheap floor, it closes the stale-read window for callbacks that arrive after deactivation starts.

client-android/serenada-core/src/main/java/app/serenada/core/call/PeerConnectionSlot.kt
```diff
@@ -618,7 +618,26 @@ internal class PeerConnectionSlot(
...
+     */
+    fun prepareTerminalClose(): Runnable? {
+        val pc = detachPeerConnectionForClose() ?: return null
+        return Runnable {
+            closePeerConnectionSafely(pc)
+            disposePeerConnectionSafely(pc, deferDispose = false)
+        }
+    }
+
+    private fun detachPeerConnectionForClose(): PeerConnection? {
```
agatx, line 639: **[P2] (confidence: 6/10) Remote audio keeps playing after `leave()` returns.**

The synchronous detach handles video thoroughly (sinks removed, tracks nulled, `onRemoteVideoTrack(remoteCid, null)`), but nothing disables the remote *audio* track or receivers, so playout continues until `pc.close()` runs on the dispose thread, behind any previously queued deferred disposes. On the slow-teardown devices this PR targets, the far end stays audible for that window while the session is already Idle and the route restore may be running concurrently. Same story for the OS mic indicator, which stays lit until the deferred ADM release.

Cheap fix: in this detach, iterate `pc.receivers` and `setEnabled(false)` (or `setVolume(0.0)`) on audio tracks before returning the native work, the audio equivalent of the sink removal above. `getReceivers()` is a fast native call, safe on Main.

client-android/serenada-core/src/main/java/app/serenada/core/call/PeerConnectionSlot.kt
```diff
@@ -1551,6 +1567,10 @@ internal class PeerConnectionDisposeQueue(
         handler.postDelayed(wrapper, delayMs)
     }
 
+    fun post(dispose: () -> Unit) {
```
agatx: **[P3] (confidence: 7/10) Two hardening nits in the queue while you're here.**

1. `post()` reads as "run on the dispose thread", but it posts to the constructor `handler` (the Main looper in production, WebRtcEngine.kt:93) with 0 delay. The runnable only reaches `serenada-pc-dispose` because `release()` calls `flush()` in the same Main-thread frame. A future caller who uses `post()` without an immediate flush gets terminal native teardown back on Main, the exact regression this PR fixes, and `PeerConnectionDisposeQueueTest` would keep passing because it always flushes. Either document the contract on `post()` or enqueue straight to `flushHandler`.
2. In `flush()`, `runnable.run()` is not wrapped in try/finally, so one throwing dispose skips `remaining.decrementAndGet()`, and `onDrained`/`shutdown()` never run: factory dispose, ADM release, and `quitSafely()` are all silently skipped. The peer teardowns and factory/ADM calls are `runCatching`-wrapped, but `releaseLocalMedia`'s `dispose()` calls are the one bare stretch in the terminal path. try/finally around `run()` plus `runCatching` on those disposes is cheap insurance.

client-android/serenada-core/src/test/java/app/serenada/core/call/PeerConnectionDisposeQueueTest.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    @Test
+    fun `flush drains posted terminal teardown on the dispose thread`() {
+        val queue = PeerConnectionDisposeQueue(Handler(Looper.getMainLooper()))
+        val drained = CountDownLatch(1)
+        val calls = mutableListOf<String>()
+        var teardownThread = ""
+        var completionThread = ""
+
+        queue.post {
```
agatx: **[P3] (confidence: 7/10) One cheap test would pin the invariant that actually prevents a native crash.**

The ordering that keeps previously deferred mid-call disposes ahead of the terminal runnable, and therefore every `PeerConnection.dispose()` ahead of `peerConnectionFactory.dispose()`, rests entirely on `LinkedHashSet` insertion order plus sequential posting in `flush()`. Nothing pins it, and the interleaving is realistic (peer replaced mid-call with `deferDispose = true`, then `leave()` before `DEFERRED_DISPOSE_DELAY_MS` elapses).

Suggested: `queue.postDelayed(a, 60_000); queue.post(b); queue.flush(...)` and assert execution order `[a, b]`. While in the area, one more assertion in `PeerConnectionSlotOnTrackMidTest`: `assertNull(h.slot.prepareTerminalClose())` after the first call, since that idempotence is what makes `engine.release()` safe after a mid-call `closePeerConnection`.

agatx (commented):
Reviewed with multiple independent passes (2x Claude subagents with fresh context, Codex cross-model, plus a manual deep-dive), all verified against the actual code before posting. Verified locally: both touched test classes pass (`:serenada-core:testDebugUnitTest`), and `node scripts/check-resilience-constants.mjs` is green (27 constants match, no shared constants touched).

**What checked out clean:** state detachment on Main is complete and idempotent (`isClosing` is `@Volatile` and gates every observer path, so late native callbacks after the deferred close are safe); dispose ordering is correct (LinkedHashSet insertion order keeps previously deferred mid-call disposes ahead of the terminal runnable, and factory/ADM release is last); `deactivateCallSession()` fully subsumes the skipped `deactivate()` for the default coordinator; mid-call replacement paths in PeerNegotiationEngine always pair `engineRemoveSlot` with `closePeerConnection(deferDispose = true)`, so dropping the session-side close loop leaks nothing; `close()` still waits for the in-flight deactivation job; the new comments and KDoc match what the code does.

**Findings (inline):**
- P1: fast leave-then-rejoin can open the camera while the old capturer is still being released on the dispose thread, silently downgrading the new call to audio-only (found independently by 3 of 4 passes)
- P1: nothing serializes the old session's worker-thread audio restore against the next session's activation, so a fast rejoin can clobber the live call's audio mode/route (or make the in-call mode sticky)
- P2: `DefaultAudioCoordinator` was Main-confined and now gets mutated from a worker with no synchronization; hang-up is exactly when proximity/device callbacks fire
- P2: remote audio keeps playing after `leave()` returns until the deferred `pc.close()` drains
- P3: `PeerConnectionDisposeQueue.post()` contract footgun + missing try/finally in `flush()`
- P3: one cheap test would pin the factory-disposed-last ordering invariant

The two P1s share one root cause: `release()`/`resetResources()` return with no completion fence, and nothing in the SDK stops the host from joining again inside the teardown window, on exactly the slow devices this PR targets. If fast rejoin is considered out of scope for ANDROID-3860, worth saying so explicitly in the PR body and tracking as a follow-up; otherwise I'd gate the next session's media/audio acquisition on the prior teardown's completion before merging. Everything else can be a follow-up.

client-android/serenada-core/src/main/java/app/serenada/core/SerenadaSession.kt
agatx: Fixed in `79d648f8` with the split suggested here. `prepareDeactivation()` runs on the owning path first: it marks the session inactive, unregisters proximity/device callbacks, clears pinned state, and snapshots the process-global values. Only the blocking `AudioManager` restore and focus abandonment move to `Dispatchers.Default`. Focus and proximity callbacks also reject inactive sessions.

client-android/serenada-core/src/main/java/app/serenada/core/call/WebRtcEngine.kt
agatx: Verified fixed in 79d648f8. `WebRtcEngine.release()` reserves a `terminalMediaTeardownFence` ticket synchronously on Main, and `startJoinInternal` awaits both fences off Main (10s bound) before audio activation and `startLocalMedia`, so the camera/ADM can no longer be re-acquired mid-teardown. Timeout now fails the join explicitly instead of silently degrading to audio-only. FIFO order is pinned by `ProcessTeardownFenceTest`. Full `:serenada-core:testDebugUnitTest` (401 tests) and `lintDebug` pass.

client-android/serenada-core/src/main/java/app/serenada/core/SerenadaSession.kt
agatx: Verified fixed in 79d648f8. The old coordinator's restore is now ordered behind a process-wide `defaultAudioTeardownFence`: `prepareDeactivation()` reserves the ticket synchronously inside `leave()` and snapshots the previous mode/route/focus before any new session can activate, and both the join path and `activateCallSession` itself await pending teardowns first. Snapshot-based restore also removes the stale-previousAudioMode corollary. The related staleness gate (phase/joinAttemptSerial re-check after activation, with compensating deactivation) is covered by the new contract test.

client-android/serenada-core/src/main/java/app/serenada/core/SerenadaSession.kt
agatx: Verified fixed in 79d648f8, using exactly the suggested split: `prepareDeactivation()` mutates all coordinator state and unregisters listeners on Main, `finishDeactivation()` runs only the blocking AudioManager restore off an immutable `Deactivation` snapshot, so the worker no longer touches shared fields. The proximity and focus listeners additionally early-return when `audioSessionActive` is false. One residual note, not blocking: `runCatching { audioCoordinator.deactivateCallSession() }` in the stale-join branch would swallow a CancellationException between `begin()` and `withContext` if that coroutine ever gains a suspension point before the check; unreachable today because providerScope is Main.immediate, just worth remembering if that flow is refactored.

client-android/serenada-core/src/main/java/app/serenada/core/call/PeerConnectionSlot.kt
agatx: Verified fixed in 79d648f8: `detachPeerConnectionForClose()` now iterates `pc.receivers` and does `setVolume(0.0)` + `setEnabled(false)` on remote audio tracks synchronously on Main, so playout stops at leave() rather than at the deferred close. Applies to mid-call replacement closes too, which is fine since those slots are being torn down.

client-android/serenada-core/src/main/java/app/serenada/core/call/PeerConnectionSlot.kt
agatx: Verified fixed in 79d648f8. `post()` is replaced by `enqueueForFlush()`, which never touches the Main handler, so the footgun is gone structurally (and it throws on a shut-down queue instead of running inline). `flush()` wraps `runnable.run()` in runCatching with try/finally accounting, onDrained/shutdown are exception-safe, and every dispose in `releaseLocalMedia` plus the peer teardowns are runCatching-wrapped. The throwing-disposal path is pinned by a new test.

client-android/serenada-core/src/test/java/app/serenada/core/call/PeerConnectionDisposeQueueTest.kt
agatx: Verified fixed in 79d648f8: `flush preserves delayed disposal before terminal teardown` pins the deferred-before-terminal (factory-disposed-last) ordering, `throwing disposal does not prevent remaining work or drain` covers the exception path, and the slot test now asserts `prepareTerminalClose()` idempotence. All 401 serenada-core unit tests pass.

agatx:
Reviewed `7d5b7f30` (custom audio coordinator fencing). Verified against the code, full `:serenada-core:testDebugUnitTest` (402 tests) and `lintDebug` pass.

The refactor is an improvement on three fronts beyond its stated goal:

1. **Custom coordinators get the same cross-session guarantee.** Moving the fence from `DefaultAudioCoordinator` into `SerenadaSession.beginAudioCoordinatorDeactivation()` closes the gap the previous fix left: a host-shared custom coordinator (the Zello case) had no ordering between the old session's `deactivateCallSession()` and the next session's activation, since the mutex is per-session. Now every coordinator teardown reserves a FIFO ticket synchronously inside `leave()`, and the interface KDoc documents the contract (activation waits for prior deactivations; deactivation must not return early and should be idempotent).
2. **Ticket reservation no longer depends on the mutex fast path.** Previously the default coordinator's ticket was reserved inside `deactivateCallSession()`, which could be delayed if leave() landed while an in-flight activation held the mutex. `begin()` now runs unconditionally before the coroutine launches, so the reservation is always synchronous with `leave()`.
3. **The stale-join branch is fixed properly.** `deactivationJob.join()` (reusing or starting the job) replaces `runCatching { deactivateCallSession() }`, which removes the CancellationException-swallow / ticket-leak hazard flagged in the previous round. Ticket completion sits in the coroutine's `finally`, which runs on every exit path including cancellation, and a scope-cancelled launch is unreachable here because `close()` only cancels the scope after the tracked job completes.

One behavioral nuance, informational: for the default coordinator, `prepareDeactivation()` (listener unregistration + `audioSessionActive` flip on Main) now runs after `awaitTurn`, where the previous commit detached immediately and fenced only the blocking restore. If a prior session's audio teardown is still pending, this coordinator's proximity/device listeners stay live for up to 10s after `leave()` and can re-apply call routing on a sensor event in that window. It needs two overlapping teardowns, is self-correcting once the deactivation runs, and is bounded by the handoff timeout, so fine to ship; just noting the trade if a field report ever shows a transient route flip during rapid call switching.

Also verified: `startJoinInternal` is single-shot through the public API (start/permission-gate/cancelJoin all route through `cleanupCall`), so the `?: beginAudioCoordinatorDeactivation()` fallback is defensive, not load-bearing; the new contract test pins the custom-coordinator FIFO behavior; and `TestSessionFactory.tearDown`'s fence drain keeps the process-wide singletons from leaking across tests (minor: its `check` throwing during cleanup can mask a test's original failure, not worth blocking on).

Ready to merge from my side.

PR #145, opened 2026-07-16
[IOS-4491] iOS: consolidate SwiftPM packages into a single root manifest

942v-zello:
Replace the per-package manifests under client-ios/ with one Package.swift at the repo root exposing SerenadaCore, SerenadaCallUI, and SerenadaBroadcastExtensionSupport. SwiftPM requires a root manifest for remote (Git URL) consumption, and a single manifest removes the duplicate maintenance of target paths and the WebRTC pin.

- SerenadaiOS consumes the root package via project.yml (path: ..)
- WebRTC pin (zello-ios-web-rtc, exact) now lives only in the root manifest; the committed xcworkspace Package.resolved remains the lockfile
- Root Package.resolved/.swiftpm/.build gitignored
- worktree-validate.sh pin check reads the root manifest + xcworkspace lockfile; docs.yml builds DocC from the repo root schemes
- Docs updated to match


Package.swift
```diff
@@ -2,12 +2,13 @@
 
 import PackageDescription
 
-// WebRTC comes from the zello-ios-web-rtc SPM package, which ships the
-// prebuilt XCFramework as a binary target. Pinned to an exact version for
-// reproducible builds; bump deliberately and re-verify call resilience.
+// Single manifest for the repo's iOS packages. External consumers depend on
```
agatx, line 5: [P1] Please migrate the remaining iOS sample and integration docs to this root package before deleting the nested manifests. `samples/ios/project.yml` still points at `../../client-ios/SerenadaCore` and `../../client-ios/SerenadaCallUI`; after `xcodegen generate`, `xcodebuild -resolvePackageDependencies` fails with exit 74 because `client-ios/SerenadaCallUI/Package.swift` no longer exists. `samples/ios/README.md` and `docs/sdk/sdk-integration-ios.md` repeat those stale paths. The sample should declare the single root `Serenada` package and select the two products.

client-ios/project.yml
```diff
@@ -16,16 +16,17 @@ packages:
...
     from: 0.0.1
-  SerenadaCore:
-    path: SerenadaCore
-  SerenadaCallUI:
-    path: SerenadaCallUI
+  # Single manifest at the repo root exposes SerenadaCore, SerenadaCallUI, and
+  # SerenadaBroadcastExtensionSupport; external consumers use the same package
+  # via this repo's Git URL.
+  Serenada:
+    path: ..
```
agatx, line 31: [P2] Could we make the generated project reference independent of the checkout directory basename? With `path: ..`, XcodeGen names the PBX file reference after the parent folder. Regenerating in `/tmp/connected-pr145-review` changed the committed `serenada` reference to `connected-pr145-review`; the main checkout is named `connected`, and normal worktrees will similarly dirty `project.pbxproj` after bootstrap. The generated project should stay stable across checkout/worktree names.

agatx (commented):
Two issues found during local validation of this head; details inline.

client-ios/project.yml
942v-zello: We don't need to commit the xcodeproj at all, we have the project.yml that regenerates it. I've deleted it and added to gitignore + updated the doc

agatx (approved):

PR #147, opened 2026-07-23
IOS-4505: Fix legacy screen share invisible to remote when started from a video-off call

danyellz:
## Summary

Legacy screen share was invisible to the remote participant when the share was started from a call that began with video off. The sharer saw their own screen locally, but the receiver rendered a blank surface even though media was flowing.

**JIRA Ticket**: [IOS-4505](https://zelloptt.atlassian.net/browse/IOS-4505)

## Root cause

Legacy screen share reuses the single camera video track, and the remote renders that surface only when the peer's advertised `videoEnabled` is true. `startScreenShareLegacy()` force-enables the local video track but never broadcasts the local media state, so the peer never learns video turned on. The manual video toggle (`toggleVideo()`) does broadcast media state, which is why enabling video before starting a share is a working manual workaround.

## Fix

- Broadcast local media state on legacy screen-share start, so the peer renders the shared surface.
- Capture the pre-share video preference and, on stop, restore it and re-broadcast media state, so a call that began audio-only returns to video-off (camera not left live) and the peer is notified.

Both stop routes (the in-app control and stopping the broadcast from the system UI) funnel through the same `onScreenShareStopped` handler, so the restore fires either way. The independent-content path is unaffected: it rides a separate content track and never preempts the camera.

## Scope

One file, `SerenadaSession.swift`. No public API changes.

## Testing

Root cause confirmed on device: starting a share from a video-off call leaves the receiver blank, and manually enabling video first (which broadcasts media state via `toggleVideo()`) makes the receiver render. This change applies that same media-state broadcast automatically on start and restores state on stop. On-device verification of the fixed build is in progress.

Surfaced while integrating screen sharing into the Zello iOS client (IOS-4505).

### DIFF
--- a/client-ios/SerenadaCore/Sources/SerenadaSession.swift
+++ b/client-ios/SerenadaCore/Sources/SerenadaSession.swift
@@ -273,6 +273,11 @@ public final class SerenadaSession: ObservableObject {
     private var cameraPermissionRequestInFlight = false
     private var isScreenShareStartPending = false
     private var screenShareStartRequestSerial: Int64 = 0
+    /// Legacy path only: the user's video preference captured at screen-share
+    /// start, restored on stop so ending a share returns video to its pre-share
+    /// state (e.g. a call that began audio-only goes back to video-off) instead
+    /// of leaving the camera live.
+    private var preScreenShareVideoPreferred = false
     private var isVideoPausedByProximity = false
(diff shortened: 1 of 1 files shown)


942v-zello (approved):
FYI: We will need to issue a new version of serenada after this merges

agatx (approved):

PR #150, opened 2026-08-20
VoIP analytics fixes

zello-tetiana:
`count_disconnects` was overcounted by one on almost every call because a dropout was counted as soon as it opened. A dropout can also happen during a normal call end when the remote peer leaves, so this case was being included in the disconnect count.

The fix moves the increment to dropout close:

Recovered dropout -> counted together with countReconnects
Peer leaves -> closed without counting a disconnect
Still disconnected at finalize() -> counted as a real disconnect

Applied across Android, iOS, and web with matching tests. Verified on a real device.

### DIFF
--- a/client-android/serenada-core/src/main/java/app/serenada/core/call/CallQualityTracker.kt
+++ b/client-android/serenada-core/src/main/java/app/serenada/core/call/CallQualityTracker.kt
@@ -136,9 +136,7 @@ internal class CallQualityTracker(
             return
         }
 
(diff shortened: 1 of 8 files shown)


agatx:
@codex please review

agatx:
@zello-tetiana Can you remove ticket references such as "ANDROID-3835" from comments

zello-tetiana:
> @zello-tetiana Can you remove ticket references such as "ANDROID-3835" from comments

@agatx Done)

